Issue #40, opened 2023-05-19
Clean up and simplify CellComplex


josefhoppe:
The CellComplex class contains some non-working, unused, or duplicate code / methods. This should be cleaned up to make it easier for both development and use.

josefhoppe:
After looking into it, I'm not sure what is useful and what isn't.

It feels a bit like a lot of the code was copied from CombinatorialComplex: For example, graph connectivity has a parameter `s` for how connected cells need to be
  - This also breaks a lot of stuff since the `s` parameter is passed to functions that don't support it
  - This also brings different implementations of adjacency matrix (one general, three seperate for node, edge, cell)
  - `CellComplex._incidence_to_adjacency()` is copied 1:1 from `CombinatorialComplex` except for `s`. Maybe generalize and move it to a utility package?

I also have some minor remarks where I'm not sure what the best way to proceed is:
- [ ] `CellComplex.incidence_matrix()` for nodes returns a $1\times n$ matrix instead of $0\times n$ (seems incorrect, but maybe something depends on this behavior?)
  - with `index=True`, it returns an empty row index list, but matrix has one row.
  - with `index=True` it also returns the indices in the wrong (inconsistent with higher ranks) order
- Interesting Property of the CellComplex Laplacian (may be correct, but unexpected): If two cells `(1,2,3,4)` and `(1,2,4,3)` exist, `(1,2)` and `(3,4)` induce the same orientation on one cell and opposite orientations on the other. This means the laplacian will have an entry of $0 (=-1 + 1)$ for these two edges.
  - [ ] In `CellComplex`, this also affects the `adjacency_matrix` as it takes the laplacian instead of using the square of the absolute of the incidence matrix. (would the correct adjacency then be 2 for twice adjacent or 1 for 'yes, adjacent'?)
- [ ] `k_hop_coincidence_matrix`: doesn't work as `coincidence_matrix` doesn't exist. (also, I'm not sure why we have it. It's just the coincidence of the k-hop (lower and upper) adjacency, isn't it? the last step seems trivial enough to leave it to the user unless the overall concept is very important)
- [ ] `get_cell_attributes`: default rank `None`, doesn't return anything and no error for rank `None` -> Easy to accidentally use incorrectly, like in `get_filtration` (see my bugfix in 0d096a46). I suggest to remove the default value.

Finally, I found some things that are straightforward bugs (I just listed them so I won't forget about it):
- [ ] `CellComplex.restrict_to_cells()` discards attributes

[...]

PR #52, opened 2023-05-22
Fix CellComplex incidence matrix and test case

josefhoppe:
From #40:

>`CellComplex.incidence_matrix()` for nodes returns a $1\times n$ matrix instead of $0\times n$ (seems incorrect, but maybe something depends on this behavior?)
> - with `index=True`, it returns an empty row index list, but matrix has one row.
> - with `index=True` it also returns the indices in the wrong (inconsistent with higher ranks) order


ninamiolane (approved):
Excellent, thanks!

Issue #59, opened 2023-05-22
Complex classes do not fulfil `Complex` contract


ffl096:
Currently the individual complex classes all miss to implement some methods that should be there by the `Complex` abstract class:

- [x] `CellComplex.clone`
- [x] `CombinatorialComplex.clone`
- [x] `SimplicialComplex.clone`
- [x] `SimplicialComplex.remove_nodes`

Additionally, `Complex` fails to derive from `abc.ABC` and thus the existing declarations of abstract methods in that class are not enforced. Otherwise it would have been more easy to spot that these implementations are missing.

PR #60, opened 2023-05-22
Fix wrong usage of `abc.abstractstaticmethod`

ffl096:
The usage of `abc.abstractstaticmethod` was wrong in the abstract complex class, as the methods are not static (and `abc.abstractstaticmethod` is deprecated anyway).

That being said, currently these decorators have no effect at the moment, as the `Complex` class does not derive from `abc.ABC`, see #59.

### DIFF
--- a/toponetx/classes/complex.py
+++ b/toponetx/classes/complex.py
@@ -39,38 +39,38 @@ class Complex:
         pass
 
     @property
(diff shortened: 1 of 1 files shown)


ninamiolane (approved):
Good eye, thanks!

PR #62, opened 2023-05-22
Update README.md

ninamiolane:


ffl096:
As per `pyproject.toml`, the minimum required python version is 3.9, not 3.10+ as the badge says. Either raise the minimum version or use the 3.9+ badge?

PR #89, opened 2023-05-23
Add missing calls to constructors of superclass

ffl096:
Even though the constructor of the superclass is currently empty, it is generally a good idea to always call it anyway. This avoids issues down the line if eventually something is added to the superclass.


ninamiolane (approved):
Good idea, thanks!

PR #91, opened 2023-05-23
Fix errors discovered by mypy

josefhoppe:
- typing > neighborhood_list_to_neighborhood_dict was called with iterable (in a different function), but list was specified -- since only an iterable was needed, I relaxed the type hint.
- same function used dictionary tried to use an undefined variable


ninamiolane (approved):
Good catch, thanks!

@josefhoppe note that you can use `pre-commit install` so that `black` and `isort` are automatically applied to each of your commit. (see readme).

PR #94, opened 2023-05-23
[BREAKING CHANGES] Clean up adjacency and related methods

josefhoppe:
As we discussed, I moved `_incidence_to_adjacency` to the utility package. I also cleaned up duplicate adjacency methods in `CellComplex`, resulting in breaking changes.

- I don't think removing `s` breaks anything as it didn't work before
- However, removing `node_ajacency_matrix` and `cell_adjacency_matrix` will inevitably break something (they could be re-added as proxy methods for `adjacency_matrix` if you prefer)
- Also, we could think about changing the behavior of `adjacency_matrix`: Currently, the default is to use the lower adjacency. For rank 0, this does not really make sense. I'm not sure what I prefer:
  - changing it to default to upper for rank 0, but lower for other ranks is the way it is used most often, but then the default is not consistent over all ranks
  - changing it to always default to upper or lower would be consistent, but then it would need to be specified more often.

### DIFF
--- a/toponetx/algorithms/spectrum.py
+++ b/toponetx/algorithms/spectrum.py
@@ -6,6 +6,7 @@ from numpy import linalg as LA
 from scipy.sparse import diags
 
 from toponetx.classes.cell_complex import CellComplex
+from toponetx.classes import CombinatorialComplex
 from toponetx.classes.simplicial_complex import SimplicialComplex
 from toponetx.datasets.mesh_complex import stanford_bunny
 
@@ -247,15 +248,15 @@ def simplicial_complex_hodge_laplacian_spectrum(
     return laplacian_spectrum(SC.hodge_laplacian_matrix(rank=rank))
 
 
-def cell_complex_adjacency_spectrum(CX: CellComplex, rank, weight="weight"):
+def cell_complex_adjacency_spectrum(CX: CellComplex, rank, up=False):
     """Return eigenvalues of the Laplacian of G.
 
     Parameters
     ----------
     matrix : scipy sparse matrix
 
-    weight : string or None, optional (default='weight')
-       If None, then each cell has weight 1.
+    up : bool, default False
+        whether to take the upper or lower adjacency
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


josefhoppe:
@ffl096 I think you're right, thanks. I'll update the code accordingly.

josefhoppe:
@ffl096 I think you're right, thanks. I'll update the code accordingly.

ninamiolane:
Thanks! A few questions @josefhoppe 
- Why: "As we discussed, I moved _incidence_to_adjacency to the utility package."?
- I think using the lower adjacency as a default is intuitive, because "adjacency" in graphs means lower adjacency.

I'm puzzled with flake8-docstrings is not picking up on the docstring formatting: docstrings should start with verb in imperative mode, which is not respected in one of your test files.

toponetx/algorithms/spectrum.py
```diff
@@ -245,14 +246,14 @@ def simplicial_complex_hodge_laplacian_spectrum(
...
-def cell_complex_adjacency_spectrum(CX: CellComplex, rank, weight="weight"):
+def cell_complex_adjacency_spectrum(CX: CellComplex, rank):
     """Return eigenvalues of the Laplacian of G.
 
     Parameters
     ----------
     matrix : scipy sparse matrix
-    weight : str or None, optional (default='weight')
-       If None, then each cell has weight 1.
+    up : bool, default False
```
ninamiolane, line 255: This docstrings does not follow the signature of the function.

josefhoppe:
> Why: "As we discussed, I moved _incidence_to_adjacency to the utility package."?

Maybe I misunderstood you. I included it in #40 as it existed in mostly the same form in both CombinatorialComplex and CellComplex and you said you generally agreed with de-duplicating code (now that I think of it, that could also be a reference to other de-duplications).

Either way, I still think that it makes sense to move it to the utility package, but if you disagree we can also revert that change.

> I think using the lower adjacency as a default is intuitive, because "adjacency" in graphs means lower adjacency.

Does it? Nodes (0-cells) are upper adjacent to each other via edges (1-cells), aren't they?

Regarding the doc strings: Thanks for catching that, I'll fix the ones you pointed out.

mhajij:
the duplication issue between many complexes is indeed a problem. NetworkX does not attach the adjacency matrix as a graph method but rather as a part of the linear algebra module :
https://github.com/networkx/networkx/blob/main/networkx/linalg/graphmatrix.py

same for the incidence matrix. I am not sure we can do that for all complexes though since the parameters might differ (CC for instance takes two arguments).

mhajij (approved):

mhajij:
lower_adjacency_matrix and upper_adjacency_matrix are the same thing as adjacency matrix and coadjacency matrix. Reason why I kept adjacency and coadjacency matrix is to be consist with graphs when the rank is zero. I am not sure if our output convention matches that of networkx's adjacnecy matrix but preferably so.

Issue #102, opened 2023-05-24
Add hints for parameters and return values!


ffl096:
TopoNetX for the most part does not use [type hints](https://docs.python.org/3/library/typing.html) for function parameters and return values, which would be usable by static type checkers such as mypy (in TopoNetX itself but also when consumers want to check _their_ code for errors).

However, recently, some comments by @josefhoppe added a few type hints to the source code, e.g., in #94. I think there should be a clear decision whether to add type hints or not:

- If not, remove the existing ones again.
- If yes, make this clear. Even if not adding them to the existing code right away, any new code should not be merged without type hints. See also [mypy's guide for existing codebases](https://mypy.readthedocs.io/en/stable/existing_code.html).

Either way, the decision should be made clear in the contribution guidelines and enforced for new pull requests.

mhajij:
what is your recommendation ?

ffl096:
I personally always use type annotations in my code, as it generally improves IDE support (suggestions, errors).

mhajij:
ok then, I will start addressing this issue then.

josefhoppe:
I agree. In the medium term, I think we should also add mypy to the linters as it can easily detect some type errors (I just ran it manually and it found two mistakes that can be easily overlooked by humans, even with the limited type hinting currently present in the codebase)

josefhoppe:
(or maybe flake8 has a type checking plugin, I don't have a strong preference for a specific tool, just type checking in general)

ffl096:
> (or maybe flake8 has a type checking plugin, I don't have a strong preference for a specific tool, just type checking in general)

flake8 doesn't do much on its own, it delegates the heavy work to other existing tools. There is [a plugin for mypy](https://pypi.org/project/flake8-mypy/) as well, but I think it's better to use mypy directly. That one really only exists as a workaround for editors that have integrated flake8 support but no integration for mypy.

michaelschaub:
It appears the discussion has a relatively clear outcome (so I changed the title of the issues). It seems, clear we should do it, but it is not a top-priority right now.  However, we should probably add this as requirement for new commits.

ninamiolane:
@ffl096 do you have updates on this? Should we close it?

ffl096:
This is an ongoing effort and will be automatically closed once #163 is merged. The remaining issues are non-trivial and will take more time.

ninamiolane:
@ffl096 when will you expect this to be completed?

[...]

Issue #96, opened 2023-05-23
Clean API for *View Python classes


ninamiolane:
# What?

The organization of the *View Python classes in not clear: 
- why is `NodeView` in `classes/node.py` but other *Views are in reportview.py? Note that `NodeView` is within reportviews in NetworkX.
- why do we use `reportview.py` whereas the sister file in `networkx` package is [reportviews.py](https://github.com/networkx/networkx/blob/main/networkx/classes/reportviews.py) (with an s)?

We need to clean this.

# Why?

The strategy behind `TopoNetX` Api and organization is to follow the `NetworkX` package as closely as possible: this gives useful guidelines and will be easier for users to get intuition about TopoNetX.

# Where?

See the files:
- [classes/node.py](https://github.com/pyt-team/TopoNetX/blob/main/toponetx/classes/node.py)
- [classes/reportview.py](https://github.com/pyt-team/TopoNetX/blob/main/toponetx/classes/reportview.py)

Or look for "View" from GitHub search bar.

# How?

- [x] Rename `reportview.py` to `reportviews.py` -- unless there was a good reason to keep it singular?
- [x] Move `NodeView` to `reportviews.py` -- unless there was a good reason to have it in its own file?
- [ ] Adapt unit-tests.

michaelschaub:
I cannot see the pull request?

mhajij:
@michaelschaub  I have added you to the review of that pull request.

mhajij:
@michaelschaub sorry it seems that we did some duplicate work :( I think I have finished addressing this issue but also did not add the unit tests for reportviews --there was an issue in the nodeview class where it had a __setitem__ method inside that class--that should not be the case since this class is a view only class. I also fixed this problem

michaelschaub:
ok closing here..

PR #103, opened 2023-05-24
adding unit tests for incidence and coincidence matrices

mhajij:


mhajij:
@josefhoppe for some reason incidence_to_adjacency(B, s=int | None= None, signed=False)  was not passing the unit tests on my local machine so I changed it to make the unit tests to incidence_to_adjacency(B, s=None, signed=False)

mhajij:
I also addressed #96

ffl096:
> @josefhoppe for some reason incidence_to_adjacency(B, s=int | None= None, signed=False) was not passing the unit tests on my local machine so I changed it to make the unit tests to incidence_to_adjacency(B, s=None, signed=False)

The union type syntax `Something | None` as equivalent replacement for `typing.Optional[Something]` is only supported since Python 3.10+. As TopoNetX supports Python 3.9+ (and I assume you are running 3.9), union type expression cannot be used. The correct way would be:
```python
from typing import Optional

# ...

def incidence_to_adjacency(B, s: Optional[int] = None, signed=False):
    ....
```

However, we should decide #102 first.

Issue #98, opened 2023-05-24
simplicial complex modes


mhajij:
In the simplicial complex constructor we have two modes : normal and ghudi. Do we need both ? 

I am not sure having both methods is useful and it is potentially harder to maintain on the long run. What do you think?

PR #106, opened 2023-05-24
Remove init mode gudhi from SimplicialComplex #98

josefhoppe:


ninamiolane (commented):
Works for me! Why that change though?

mhajij (approved):

mhajij:
@ninamiolane it is not really needed and it does not add any performance advantage--also it makes the API different from other complexes

PR #108, opened 2023-05-24
sample fix for o(n!) of combinatorial complex

bahramiramazan:
@mhajij  Combinatorial_complex has  an attribute named ' _aux_complex ' , that is an object of simplicial_complex. Adding a cell into the  combinatorial_complex , basically involves creating a simplex , and then  passing it to add_cell method of '_aux_complex'. I had to change all the three in order to add a cell of order 17 or above .However  for combinatorial complex  the coadjacency does not work after modification, and for sure it have consequences for simplicial_complex.

### DIFF
--- a/test/classes/test_simplicial_complex.py
+++ b/test/classes/test_simplicial_complex.py
@@ -191,6 +191,18 @@ class TestSimplicialComplex(unittest.TestCase):
         assert SC.maxdim == 0
         assert SC[9]["is_maximal"] is True
 
+    def test_add_node(self):
+        """Test add node."""
+        SC = SimplicialComplex()
+        SC.add_node(9)
+        assert 9 in SC
+        with self.assertRaises(ValueError):
+            SC.add_node((1, 2))
+
+        with self.assertRaises(ValueError):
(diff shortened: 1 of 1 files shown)


toponetx/classes/simplex.py
```diff
@@ -47,8 +47,8 @@ def __init__(self, elements, name=None, construct_tree=True, **attr):
         if len(self.nodes) != len(elements):
             raise ValueError("A simplex cannot contain duplicate nodes.")
 
-        if construct_tree:
-            self._faces = self.construct_simplex_tree(elements)
+        # if construct_tree:
```
ninamiolane, line 50: Remove commented code.

mhajij:
@bahramiramazan  since the problem you raised above is in the combinatorialcomplex class, then why are you modifying the simplicialcomplex class?

ninamiolane:
@bahramiramazan all checks have failed (see above): they would need to pass before we can merge. Thanks!

bahramiramazan:
> @bahramiramazan all checks have failed (see above): they would need to pass before we can merge. Thanks!

well noted ! thanks

Issue #95, opened 2023-05-23
Clean docstrings and API


ninamiolane:
# What?

The names of functions, files, folders, variables, parameters needs a clean-up. For example:
- rank is sometimes refered as `r`, sometimes as `rank`,
- docstrings do not match the functions' signatures,
- a cell complex is abbreviated as "CX"; however in the papers Hajij 2023 and Papillon 2023 we use the abbreviation CC
- a combinatorial complex is abbreviated as "CC"; however in the papers Hajij 2023 and Papillon 2023 we use the abbreviation CCC
- etc

The docstrings do not always respect our [docstring conventions](https://pyt-team.github.io/toponetx/contributing/index.html#write-documentation), which renders poorly in the [documentation website](https://pyt-team.github.io/toponetx/index.html).

# How?

Clean docstrings in `toponetx/algorithms`:
- [ ] Check that they respect the correct formatting
- [ ] Check that the Parameters description match the function signature

Clean docstrings in `toponetx/classes`:
- [ ] Check that they respect the correct formatting
- [ ] Check that the Parameters description match the function signature

Clean docstrings in `toponetx/utils`:
- [ ] Check that they respect the correct formatting
- [ ] Check that the Parameters description match the function signature

Decide on abbreviation and apply changes:
- [x] Use CC instead of CX for cell complex, everywhere
- [x] Use CCC instead of CC for combinatorial complex, everywhere

Make sure that variable/parameter names are consistent:
- [x] Use `rank` for rank everywhere

mhajij:
I started addressing this a little bit. I think the docstrings for algorithms is fine. No?

mhajij:
@devendragovil 
@maneelusf 
@gurug-dev 

This is also a priority task. Can you please help?

ninamiolane:
@ffl096 this is stalled issue. I have seen some PRs from you that were addressing it in the past. Could you check this issue's status? Maybe we can close it now? 🤞 Thanks!

ffl096:
There are unfortunately still places that need to be cleaned up, though there was some progress in the past. Let's keep this open for now.

ninamiolane:
@ffl096 would you have any update on this? Maybe we can close?

PR #116, opened 2023-05-25
WIP Improve docstrings / type hints in CellComplex

josefhoppe:
I've started improving docstrings in CellComplex (#95) and improved type hints (#102) along the way.

### DIFF
--- a/toponetx/classes/cell_complex.py
+++ b/toponetx/classes/cell_complex.py
@@ -392,7 +392,7 @@ class CellComplex(Complex):
         int
             Number of cells of rank at least rank that contain node.
         """
-        if rank >= 1:
+        if rank > 1:
             raise NotImplementedError(
                 f"Rank {rank} is currently not supported by degree."
             )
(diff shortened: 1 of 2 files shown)


ninamiolane (approved):
Excellent, thanks!

toponetx/classes/cell_complex.py
```diff
@@ -11,12 +11,14 @@
 from collections import defaultdict
 from collections.abc import Hashable, Iterable
 from itertools import pairwise, zip_longest
+from typing import Any, Dict, Optional, Set, Tuple
```
ffl096, line 14: TopoNetX is Python 3.9+, so you can use the native `dict`, `set` and `tuple` classes with subscripting. The dedicated classes in `typing` have been deprecated.

toponetx/classes/cell_complex.py
```diff
@@ -270,7 +272,7 @@ def __getitem__(self, node):
         """
         return self.neighbors(node)
 
-    def _insert_cell(self, cell, **attr):
+    def _insert_cell(self, cell: tuple | list | Cell, **attr):
```
ffl096, line 275: TopoNetX is Python 3.9+, so you cannot use union types using the pipe character (Python 3.10+ only). Use `typing.Union`.

toponetx/classes/cell_complex.py
josefhoppe: Oh, my bad. I thought those were still up-to-date, thank you.

PR #122, opened 2023-05-27
more unit tests for sc

mhajij:


papamarkou:
Thank you very much, @ninamiolane. Some more work is required in this tutorial. Here is the to-do list to be completed:

- [ ] We mention that the determination of sign in incidence matrices and up-Laplacians is carried out anticlockwise. This is not clear enough, we need to explain that the signs are based on a convention, namely on a lexicographic orientation of faces and edges. Furthermore, we need to explain how we traverse such elements. I attached the relevant hand-written note of mine that clarifies how this is carried out for example 2 of the tutorial on simplicial complexes.
- [ ] As @mhajij pointed out, in practice, `TopoNetX` uses matrix multiplication between incidence matrices for the calculation of up-Laplacians. Essentially, we use equation 7 from [this paper](https://qmro.qmul.ac.uk/xmlui/bitstream/handle/123456789/75450/Bianconi%20Spectral%20Detection%20of%20Simplicial%20Communities%20via%20Hodge%20Laplacians%202021%20Accepted.pdf?sequence=2&isAllowed=y), whereas my explanation above is based on equation 9 from the same paper.
- [ ] We need to elaborate in a similar fashion on the calculation of down-Laplacians.
- [ ] It would be good to provide a simple toy example with some real interpretation when we explain how to populate simplicial complexes with data.

I made these notes here both for myself and for anyone else who may be faster than me to attend to these matters :)
[up_laplacian.pdf](https://github.com/pyt-team/TopoNetX/files/11601307/up_laplacian.pdf)
up_laplacian

papamarkou:
@ninamiolane by mistake I added a comment to the wrong PR, I deleted it from here and moved it to the correct PR :)

PR #123, opened 2023-05-27
Fix typing annotations for Python 3.9

ffl096:
The pull request #116 introduced type annotations that are not compatible with Python 3.9 (our minimum supported version), in particular union type expressions are only available since Python 3.10. Replace them with `typing.Union`.

In addition, this pull request uses subscripting on the build-in `dict`, `list`, `set` and `tuple` classes, which is available since Python 3.9. The respective classes inside the `typing` module are no longer necessary and have been deprecated.

### DIFF
--- a/toponetx/classes/cell_complex.py
+++ b/toponetx/classes/cell_complex.py
@@ -11,7 +11,7 @@ import warnings
 from collections import defaultdict
 from collections.abc import Hashable, Iterable
 from itertools import pairwise, zip_longest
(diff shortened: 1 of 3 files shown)


ffl096:
By the way, the test suite should be executed on the latest release of every supported Python version (i.e., 3.9.x, 3.10.x and 3.11.x). This would have catched this error as tests would have failed on Python 3.9.

PR #126, opened 2023-05-30
Fix installation instruction for zsh

### LINKED ISSUE #125: Pip install -e doesn't work
I was following the steps in "Getting Started" section of README file, and the "pip install -e [.all]" command doesn't seem to work for me. 

I was able to resolve it by first upgrading the PIP version (it seems that a[ pip >= 21.3 is required](https://stackoverflow.com/questions/69711606/how-to-install-a-package-using-pip-in-editable-mode-with-pyproject-toml)). 

I am also using zsh and zsh seems to require a single quote around [.all]. So this works `pip` install -e '[.all]'\

ffl096:
Closes #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ cd TopoNetX
 ```bash
 git pull
 ```
-3. Install `TopoNetX` in editable mode:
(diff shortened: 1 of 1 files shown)


ninamiolane (approved):
Thanks!!

PR #128, opened 2023-05-31
Run test suite on latest release of all supported python versions

ffl096:
With this change, our tests are run once for each supported Python version. For the future, this should prevent errors as fixed by #123 before they are merged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.10.11]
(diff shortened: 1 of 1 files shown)


ffl096:
The failing tests of this pull request again highlights why we should add this. The error is in existing code, as `itertools.pairwise` is only available since Python 3.10.

josefhoppe (commented):
Good idea, I'll submit a separate PR to remove the errors. I don't think we should merge this PR before that is done because in the meantime, the fact that tests are always failing may confuse people submitting PRs.

josefhoppe (approved):

PR #131, opened 2023-05-31
Fixed a name

papamarkou:


toponetx/datasets/mesh.py
```diff
@@ -12,14 +12,14 @@
 DIR = os.path.dirname(__file__)
 
 
-def stanford_bunny(cmplex_type="simplicial complex"):
+def stanford_bunny(domain="simplicial"):
```
ninamiolane, line 15: Using the `domain` name is a great idea, thank you!

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,28 @@
+from torch_geometric.utils.convert import to_networkx
```
ninamiolane, line 1: We should not depend on torch_geometric, because users of toponetx do not want to download torch. instead, we can use another data source for the karate club graph.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,28 @@
...
+            networkx_graph, max_dim=max_dim
+        )
+
+    # def to_cellular(self, directed):
+    #     pass
+
+
+# dataset = GraphDataset(KarateClub(), directed=True, domain="cellular")
+# dataset.graph
+# dataset.domain
```
ninamiolane, line 28: The most important part is: create a default method that generates features on edges, faces and higher order cells of the new domain.

ninamiolane (commented):
thanks!

toponetx/datasets/mesh.py
ninamiolane: Also love the idea of shortening to "simplicial"

PR #134, opened 2023-06-01
Shrec 16 dataset

mhajij:
The branch is not unit tested for the following reason  :: 

(1) Unable to push large files to the repo--file size is restricted.
(2) Unable to fetch these large files from a different url, The large files are provided in the repo at the end 

The main function written here is shrec_16()

large files : https://github.com/mhajij/shrec_16

### DIFF
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -1,0 +1,29 @@
+"""Test mesh dataset."""
+
+import unittest
+
+from toponetx import CellComplex, SimplicialComplex
+from toponetx.datasets.graph import karate_club_complex
+
+
+class TestMeshDatasets(unittest.TestCase):
+    """Test datasets utils."""
+
+    def test_karate_club_complex(self):
+        """Test stanford_bunny."""
+        simplicial_karate_club_data = karate_club_complex("simplicial complex")
+
+        assert "node_feat" in simplicial_karate_club_data
+        assert "edge_feat" in simplicial_karate_club_data
+        assert "face_feat" in simplicial_karate_club_data
+        assert "tet_feat" in simplicial_karate_club_data
+
+        cell_karate_club_data = karate_club_complex("cell complex")
+
+        assert "node_feat" in cell_karate_club_data
+        assert "edge_feat" in cell_karate_club_data
+        assert "cell_feat" in cell_karate_club_data
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -8,7 +8,7 @@ class TestGraph:
     """Test datasets utils."""
 
     def test_karate_club(self):
-        """Test stanford_bunny."""
+        """Test karate_club."""
         simplicial_karate_club_data = karate_club("simplicial")
 
         assert "node_feat" in simplicial_karate_club_data
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -14,7 +14,7 @@ class TestGraph:
         assert "node_feat" in simplicial_karate_club_data
         assert "edge_feat" in simplicial_karate_club_data
         assert "face_feat" in simplicial_karate_club_data
-        assert "tet_feat" in simplicial_karate_club_data
+        assert "tetrahedron_feat" in simplicial_karate_club_data
 
         cell_karate_club_data = karate_club("cell")
 
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -11,14 +11,14 @@ class TestMeshDatasets(unittest.TestCase):
 
     def test_karate_club_complex(self):
         """Test stanford_bunny."""
-        simplicial_karate_club_data = karate_club_complex("simplicial complex")
+        simplicial_karate_club_data = karate_club_complex("simplicial")
 
         assert "node_feat" in simplicial_karate_club_data
         assert "edge_feat" in simplicial_karate_club_data
         assert "face_feat" in simplicial_karate_club_data
         assert "tet_feat" in simplicial_karate_club_data
 
-        cell_karate_club_data = karate_club_complex("cell complex")
+        cell_karate_club_data = karate_club_complex("cell")
 
         assert "node_feat" in cell_karate_club_data
         assert "edge_feat" in cell_karate_club_data
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -11,13 +11,16 @@ class TestGraph:
         """Test karate_club."""
         simplicial_karate_club_data = karate_club("simplicial")
 
-        assert "node_feat" in simplicial_karate_club_data
-        assert "edge_feat" in simplicial_karate_club_data
-        assert "face_feat" in simplicial_karate_club_data
-        assert "tetrahedron_feat" in simplicial_karate_club_data
+        assert len(simplicial_karate_club_data.get_simplex_attributes("node_feat")) != 0
+        assert len(simplicial_karate_club_data.get_simplex_attributes("edge_feat")) != 0
+        assert len(simplicial_karate_club_data.get_simplex_attributes("face_feat")) != 0
+        assert (
+            len(simplicial_karate_club_data.get_simplex_attributes("tetrahedron_feat"))
+            != 0
+        )
 
         cell_karate_club_data = karate_club("cell")
 
-        assert "node_feat" in cell_karate_club_data
-        assert "edge_feat" in cell_karate_club_data
-        assert "cell_feat" in cell_karate_club_data
+        assert len(cell_karate_club_data.get_cell_attributes("node_feat", rank=0)) != 0
+        assert len(cell_karate_club_data.get_cell_attributes("edge_feat", rank=1)) != 0
+        assert len(cell_karate_club_data.get_cell_attributes("cell_feat", rank=2)) != 0
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -1,4 +1,4 @@
-"""Test mesh dataset."""
+"""Test graph dataset."""
 
 import unittest
 
@@ -6,7 +6,7 @@ from toponetx import CellComplex, SimplicialComplex
 from toponetx.datasets.graph import karate_club
 
 
-class TestMeshDatasets(unittest.TestCase):
+class TestGraph(unittest.TestCase):
     """Test datasets utils."""
 
     def test_karate_club(self):
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -9,7 +9,7 @@ from toponetx.datasets.graph import karate_club_complex
 class TestMeshDatasets(unittest.TestCase):
     """Test datasets utils."""
 
-    def test_karate_club_complex(self):
+    def test_karate_club(self):
         """Test stanford_bunny."""
         simplicial_karate_club_data = karate_club_complex("simplicial")
 
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -3,7 +3,7 @@
 import unittest
 
 from toponetx import CellComplex, SimplicialComplex
-from toponetx.datasets.graph import karate_club_complex
+from toponetx.datasets.graph import karate_club
 
 
 class TestMeshDatasets(unittest.TestCase):
@@ -11,14 +11,14 @@ class TestMeshDatasets(unittest.TestCase):
 
     def test_karate_club(self):
         """Test stanford_bunny."""
-        simplicial_karate_club_data = karate_club_complex("simplicial")
+        simplicial_karate_club_data = karate_club("simplicial")
 
         assert "node_feat" in simplicial_karate_club_data
         assert "edge_feat" in simplicial_karate_club_data
         assert "face_feat" in simplicial_karate_club_data
         assert "tet_feat" in simplicial_karate_club_data
 
-        cell_karate_club_data = karate_club_complex("cell")
+        cell_karate_club_data = karate_club("cell")
 
         assert "node_feat" in cell_karate_club_data
         assert "edge_feat" in cell_karate_club_data
--- a/test/datasets/test_graph.py
+++ b/test/datasets/test_graph.py
@@ -1,12 +1,10 @@
 """Test graph dataset."""
 
-import unittest
-
 from toponetx import CellComplex, SimplicialComplex
 from toponetx.datasets.graph import karate_club
 
 
-class TestGraph(unittest.TestCase):
+class TestGraph:
     """Test datasets utils."""
 
     def test_karate_club(self):
@@ -23,7 +21,3 @@ class TestGraph(unittest.TestCase):
         assert "node_feat" in cell_karate_club_data
         assert "edge_feat" in cell_karate_club_data
         assert "cell_feat" in cell_karate_club_data
-
-
-if __name__ == "__main__":
-    unittest.main()
--- a/test/transform/test_graph_to_cell_complex.py
+++ b/test/transform/test_graph_to_cell_complex.py
@@ -1,20 +1,15 @@
-"""Test graph to simplicial complex transformation."""
+"""Test graph to cell complex transformation."""
 
-import unittest
 
 import networkx as nx
 
 from toponetx.transform.graph_to_cell_complex import homology_cycle_cell_complex
 
 
-class TestGraphToCellComplex(unittest.TestCase):
-    """Test graph to simplicial complex transformation."""
+class TestGraphToCellComplex:
+    """Test graph to cell complex transformation."""
 
     def test_homology_cycle_cell_complex(self):
         """Test homology_cycle_cell_complex."""
         cx = homology_cycle_cell_complex(nx.karate_club_graph())
         assert len(cx.cells) != 0
-
-
-if __name__ == "__main__":
-    unittest.main()
--- a/test/transform/test_graph_to_cell_complex.py
+++ b/test/transform/test_graph_to_cell_complex.py
@@ -1,0 +1,20 @@
+"""Test graph to simplicial complex transformation."""
+
+import unittest
+
+import networkx as nx
+
+from toponetx.transform.graph_to_cell_complex import homology_cycle_cell_complex
+
+
+class TestGraphToCellComplex(unittest.TestCase):
+    """Test graph to simplicial complex transformation."""
+
+    def test_homology_cycle_cell_complex(self):
+        """Test homology_cycle_cell_complex."""
+        cx = homology_cycle_cell_complex(nx.karate_club_graph())
+        assert len(cx.cells) != 0
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/toponetx/__init__.py
+++ b/toponetx/__init__.py
@@ -19,7 +19,7 @@ from .classes.reportviews import CellView, HyperEdgeView, SimplexView, NodeView
 
 
 from .datasets.mesh import stanford_bunny, shrec_16
-from .datasets.graph import karate_club_complex
+from .datasets.graph import karate_club
 
 from .transform.graph_to_simplicial_complex import (
     graph_2_neighbor_complex,
--- a/toponetx/__init__.py
+++ b/toponetx/__init__.py
@@ -18,12 +18,16 @@ from .classes.cell import Cell
 from .classes.reportviews import CellView, HyperEdgeView, SimplexView, NodeView
 
 
-from .datasets.mesh import stanford_bunny
+from .datasets.mesh import stanford_bunny, shrec_16
+from .datasets.graph import karate_club_complex
 
 from .transform.graph_to_simplicial_complex import (
     graph_2_neighbor_complex,
     graph_2_clique_complex,
 )
+
+from .transform.graph_to_cell_complex import homology_cycle_cell_complex
+
 
 from .utils.structure import (
     sparse_array_to_neighborhood_list,
--- a/toponetx/__init__.py
+++ b/toponetx/__init__.py
@@ -15,7 +15,7 @@ from .classes.reportviews import CellView, HyperEdgeView, NodeView, SimplexView
 from .classes.simplex import Simplex
 from .classes.simplicial_complex import SimplicialComplex
 from .datasets.graph import karate_club
-from .datasets.mesh import stanford_bunny, shrec_16
+from .datasets.mesh import shrec_16, stanford_bunny
 from .transform.graph_to_cell_complex import homology_cycle_cell_complex
 from .transform.graph_to_simplicial_complex import (
     graph_2_clique_complex,
--- a/toponetx/algorithms/spectrum.py
+++ b/toponetx/algorithms/spectrum.py
@@ -117,7 +117,7 @@ def set_hodge_laplacian_eigenvector_attrs(
     Examples
     --------
     >>> from toponetx import SimplicialComplex
-    >>> SC=SimplicialComplex([[1, 2, 3],[2, 3, 5],[0, 1]])
+    >>> SC = SimplicialComplex([[1, 2, 3],[2, 3, 5],[0, 1]])
     >>> set_hodge_laplacian_eigenvector_attrs(SC, 1, 2, "down")
     >>> SC.get_simplex_attributes("0.th_eigen", 1)
     """
--- a/toponetx/algorithms/spectrum.py
+++ b/toponetx/algorithms/spectrum.py
@@ -1,14 +1,15 @@
 """Module to compute spectra."""
 
 import numpy as np
+import scipy as sp
 import scipy.sparse as sparse
+import spharapy.spharabasis as sb
 from numpy import linalg as LA
 from scipy.sparse import diags
 
 from toponetx.classes import CombinatorialComplex
 from toponetx.classes.cell_complex import CellComplex
 from toponetx.classes.simplicial_complex import SimplicialComplex
-from toponetx.datasets.mesh import stanford_bunny
 
 __all__ = [
     "hodge_laplacian_eigenvectors",
@@ -41,7 +42,6 @@ def _normalize(f):
     maxf = max(f.values())
     f_normalized = {}
     for v in f.keys():
-
         if minf == maxf:
             f_normalized[v] = 0
         else:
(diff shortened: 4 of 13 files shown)


mhajij:
@ninamiolane  although this passes all unit tests, but the funciton added is not unit tested because of failure of adding the dataset files.

josefhoppe:
I see the problem with the files being too large. However, like this, it is hard to use. (Or did you have a way to make it easier to use in mind?)

Can we either:

1. unzip the .npz files (-> all individual files are < 100 MB, so should be allowed by github; but still large)
2. For publishing to pypi, we could move it to a dedicated package that depends on toponetx -> only one package that is so large
3. download the files on-demand
4. at least add a 'how to' to the docstring and change the location (users won't want to add files to our package and it shouldn't be expected of them)

papamarkou:
One issue is that the data files are big. One more issue is that by loading features from a file, it is not clear to the users how the features are generated. I would suggest we load the mesh specification from a file, and we add re-usable methods for loading features. This way, the users will know what the features are, and they will also be able to use our functionality for generating the vertex/edge/face features across different mesh datasets.

toponetx/datasets/mesh.py
```diff
@@ -2,17 +2,20 @@
 
 import os
 import os.path
+import urllib.request
```
josefhoppe, line 5:
```suggestion
```
Not in use currently

toponetx/datasets/mesh.py
```diff
@@ -32,11 +35,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
...
+
+
+def shrec_16():
+    """Load training/testing shrec 16 datasets".
+
+    Returns
+    -------
+    tuple of length 2 npz files
+        The npz files store the training/testing complexes of shrec 16 dataset along
+        with thier nodes, edges and faces features.
```
josefhoppe, line 55:
```suggestion
        with their nodes, edges and faces features.
```
(typo)

toponetx/datasets/mesh.py
```diff
@@ -32,11 +35,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
...
+    if os.path.isfile(DIR + "/shrec_training.npz") and os.path.isfile(
+        DIR + "/shrec_testing.npz"
+    ):
+        print("Loading dataset..")
+        shrec_training = np.load(DIR + "/shrec_training.npz", allow_pickle=True)
+        shrec_testing = np.load(DIR + "/shrec_testing.npz", allow_pickle=True)
+        print("done!")
+        return shrec_training, shrec_testing
+    else:
+        raise ValueError("Files are on the HD, fail to load the dataset.")
```
josefhoppe, line 81:
```suggestion
        raise ValueError(f"Files couldn't be found in folder {DIR}, fail to load the dataset.")
```
Error message is hard to understand.

josefhoppe (changes requested):

mhajij:
> I see the problem with the files being too large. However, like this, it is hard to use. (Or did you have a way to make it easier to use in mind?)
> 
> Can we either:
> 
> 1. enable LFS for this repository
> 2. unzip the .npz files (-> all individual files are < 100 MB, so should be allowed by github; but still large)
> 3. For publishing to pypi, we could move it to a dedicated package that depends on toponetx -> only one package that is so large
> 4. download the files on-demand
> 5. at least add a 'how to' to the docstring and change the location (users won't want to add files to our package and it shouldn't be expected of them)

I think the best solution is 4. I am trying to write a script for that but it is not downloading the files correctly.

Here is what I have:

url ='https://github.com/mhajij/shrec_16/blob/main/shrec_testing.zip'


from tqdm import tqdm
import requests

response = requests.get(url, stream=True)

with open("test.zip", "wb") as handle:
    for data in tqdm(response.iter_content()):
        handle.write(data)


it is now downloading the file correctly. Any suggestions?

mhajij:
> However, that took 2 minutes for me, maybe tqdm is not a good idea? The code snippet from the question here https://stackoverflow.com/questions/16694907/download-large-file-in-python-with-requests took 2 seconds.

which code snippet?

ffl096:
@josefhoppe That has nothing to do with `tqdm`.
@mhajij Just set the `chunk_size` in `response.iter_content()` to something reasonable. The default value of `1` is just bad. Some small multiple of `1024` should be fine.

toponetx/datasets/mesh.py
```diff
@@ -32,11 +34,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
     ValueError
         If cmplex_type is not one of the supported values.
     """
-    if cmplex_type == "simplicial complex":
+    if complex_type == "simplicial complex":
         cpx = SimplicialComplex.load_mesh(DIR + "/bunny.obj")
         return cpx
-    elif cmplex_type == "cell complex":
+    elif complex_type == "cell complex":
```
ninamiolane, line 40:
```suggestion
    if complex_type == "cell complex":
```

toponetx/datasets/mesh.py
```diff
@@ -32,11 +34,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
...
         cpx = CellComplex.load_mesh(DIR + "/bunny.obj")
         return cpx
     else:
```
ninamiolane, line 43:
```suggestion
```

toponetx/datasets/mesh.py
```diff
@@ -32,11 +34,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
...
         raise ValueError("cmplex_type must be 'simplicial complex' or 'cell complex'")
```
ninamiolane, line 44:
```suggestion
    raise ValueError("cmplex_type must be 'simplicial complex' or 'cell complex'")
```

ninamiolane, line 44:
```suggestion
        raise ValueError("complex_type must be 'simplicial' or 'cell'")
```

toponetx/datasets/mesh.py
```diff
@@ -32,11 +34,47 @@ def stanford_bunny(cmplex_type="simplicial complex"):
     ValueError
         If cmplex_type is not one of the supported values.
```
ninamiolane, line 35:
```suggestion
        If complex_type is not one of the supported values.
```

toponetx/datasets/mesh.py
```diff
@@ -3,16 +3,18 @@
...
 import os.path
 
+import numpy as np
+
 from toponetx import CellComplex, CombinatorialComplex, SimplicialComplex
 
-__all__ = ["stanford_bunny"]
+__all__ = ["stanford_bunny", "shrec_16"]
 
 #: the absolute path repr the directory containing this file
```
ninamiolane, line 12:
```suggestion
```

toponetx/datasets/mesh.py
```diff
@@ -3,16 +3,18 @@
...
 # DIR = os.path.abspath(os.getcwd())
```
ninamiolane, line 13:
```suggestion
```

ninamiolane (commented):
Minor comments.

In addition, use `os.path.join(DIR, "shrec_training.npz")` instead of + with strings

ninamiolane (changes requested):
Minor comments.

In addition, use `os.path.join(DIR, "shrec_training.npz")` instead of + with strings

ninamiolane (changes requested):
Minor comments.

In addition, use `os.path.join(DIR, "shrec_training.npz")` instead of + with strings

toponetx/datasets/mesh.py
ninamiolane: Additionally, let the complex_type be one of:
- "simplicial"
- "cell"
- etc
i.e. avoid to have the " complex" part of the string
--> it's shorter and does not use spaces.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,78 @@
...
+            "complex": cx,
+            "node_feat": np.array(nodes_feat),
+            "edge_feat": np.array(edges_feat),
+            "cell_feat": np.array(cells_feat),
+        }
+        return data
+
+    else:
+        raise ValueError(
+            f"cmplex_type must be 'simplicial' or 'cell' got {complex_type}"
```
ninamiolane, line 77:
```suggestion
            f"complex_type must be 'simplicial' or 'cell' got {complex_type}"
```

toponetx/transform/graph_to_cell_complex.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Parameters
+    ----------
+    G : networkx graph
+        Input graph.
+
+    Returns
+    -------
+    CellComplex
+        the cell complex obtained by adding the homology cycles of the graph
```
ninamiolane, line 24: This output description does not respect the docstring format.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,78 @@
...
+from toponetx.algorithms.spectrum import (
+    hodge_laplacian_eigenvectors,
+    set_hodge_laplacian_eigenvector_attrs,
+)
+from toponetx.transform.graph_to_simplicial_complex import graph_2_clique_complex
+
+__all__ = ["karate_club_complex"]
+
+
+def karate_club_complex(complex_type="simplicial"):
```
ninamiolane, line 16: Rename to `karate_club` only, as the other data functions do not use the word _complex.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,78 @@
...
+        data = {
+            "complex": sc,
+            "node_feat": np.array(nodes_feat),
+            "edge_feat": np.array(edges_feat),
+            "face_feat": np.array(faces_feat),
+            "tet_feat": np.array(tet_feat),
+        }
+        return data
+
+    elif complex_type == "cell":
```
ninamiolane, line 57: No need for "elif" because the "if" statement ends with "return".
Use "if" instead.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,78 @@
...
+
+        data = {
+            "complex": cx,
+            "node_feat": np.array(nodes_feat),
+            "edge_feat": np.array(edges_feat),
+            "cell_feat": np.array(cells_feat),
+        }
+        return data
+
+    else:
```
ninamiolane, line 75: Ditto, no need for "else" because the above two if-statements end with "return". Remove else.

toponetx/transform/graph_to_cell_complex.py
```diff
@@ -0,0 +1,29 @@
...
+    "homology_cycle_cell_complex",
+]
+
+
+import networkx as nx
+
+from toponetx import CellComplex
+
+
+def homology_cycle_cell_complex(G):
```
ninamiolane, line 13: This function needs a unit-test

toponetx/datasets/mesh.py
```diff
@@ -36,14 +35,14 @@ def stanford_bunny(complex_type="simplicial complex"):
     ValueError
         If complex_type is not one of the supported values.
     """
-    if complex_type == "simplicial complex":
+    if complex_type == "simplicial":
         cpx = SimplicialComplex.load_mesh(DIR + "/bunny.obj")
         return cpx
-    elif complex_type == "cell complex":
+    elif complex_type == "cell":
```
ninamiolane, line 41: Do not use "cpx" for complex, write "complex" instead because it is not much longer and it helps code readability for contributors.

Also, replace "elif" by "if" because: the if-statement above ends with a "return", thus elif is not needed.

toponetx/datasets/mesh.py
```diff
@@ -36,14 +35,14 @@ def stanford_bunny(complex_type="simplicial complex"):
...
         cpx = CellComplex.load_mesh(DIR + "/bunny.obj")
         return cpx
     else:
```
ninamiolane, line 44: No need for else because the above two if-statements end with "return". Remove.

toponetx/transform/graph_to_cell_complex.py
```diff
@@ -0,0 +1,29 @@
...
+    G : networkx graph
+        Input graph.
+
+    Returns
+    -------
+    CellComplex
+        the cell complex obtained by adding the homology cycles of the graph
+    """
+    cycles = nx.cycle_basis(G)
+    cx = CellComplex(G)  # lift to graph
```
ninamiolane, line 27: Remove unnecessary comments: comments are hard to maintain because they are not unittested and they can become outdated. the code should be clean enough to be read without comments. if documentation is needed, then explanations should go in the docstrings.

ninamiolane (changes requested):

toponetx/datasets/graph.py
ninamiolane: Also, maybe let's rename "complex_type" to "domain" as was suggested by @papamarkou ?

toponetx/datasets/mesh.py
```diff
@@ -36,14 +35,14 @@ def stanford_bunny(complex_type="simplicial complex"):
     ValueError
         If complex_type is not one of the supported values.
     """
-    if complex_type == "simplicial complex":
+    if complex_type == "simplicial":
         cpx = SimplicialComplex.load_mesh(DIR + "/bunny.obj")
```
ninamiolane, line 39: This needs to be os.path.join(DIR, "bunny.obj")

toponetx/algorithms/spectrum.py
```diff
@@ -118,7 +118,7 @@ def set_hodge_laplacian_eigenvector_attrs(
     --------
     >>> from toponetx import SimplicialComplex
     >>> SC=SimplicialComplex([[1,2,3],[2,3,5],[0,1]])
-    >>> SC = set_hodge_laplacian_eigenvector_attrs(SC,1,2,"down")
+    >>> set_hodge_laplacian_eigenvector_attrs(SC,1,2,"down")
```
ninamiolane, line 121: Try to respect coding style guidelines in the docstrings as well, e.g. we need spaces after `,` here.

toponetx/datasets/mesh.py
```diff
@@ -36,14 +35,14 @@ def stanford_bunny(complex_type="simplicial complex"):
...
         return cpx
-    elif complex_type == "cell complex":
+    elif complex_type == "cell":
         cpx = CellComplex.load_mesh(DIR + "/bunny.obj")
```
ninamiolane, line 42: Ditto

toponetx/datasets/mesh.py
mhajij: what if the complex is not of the types the user input? that is the purpose of the third one, no?

toponetx/datasets/mesh.py
mhajij: I am not using complex because it is a keyword in Python. for complex numbers.

mhajij: I am not sure about your second comment, the else statement seems to be needed for me to take care of invalid input other than simplicial or cell, no?

toponetx/datasets/graph.py
mhajij: on my local branch this seem fine-- not sure what is going on here.

toponetx/datasets/graph.py
mhajij: what if we did not go inside the if/ifelse condition though?

test/transform/test_graph_to_cell_complex.py
```diff
@@ -0,0 +1,20 @@
+"""Test graph to simplicial complex transformation."""
+
+import unittest
```
ninamiolane, line 3: No need for `unittest` anymore, see `test/base/test_conv.py` for example.

test/transform/test_graph_to_cell_complex.py
```diff
@@ -0,0 +1,20 @@
...
+
+import networkx as nx
+
+from toponetx.transform.graph_to_cell_complex import homology_cycle_cell_complex
+
+
+class TestGraphToCellComplex(unittest.TestCase):
```
ninamiolane, line 10:
```suggestion
class TestGraphToCellComplex:
```

ninamiolane: ^^^no need for unittest

test/transform/test_graph_to_cell_complex.py
```diff
@@ -0,0 +1,20 @@
...
+    """Test graph to simplicial complex transformation."""
+
+    def test_homology_cycle_cell_complex(self):
+        """Test homology_cycle_cell_complex."""
+        cx = homology_cycle_cell_complex(nx.karate_club_graph())
+        assert len(cx.cells) != 0
+
+
+if __name__ == "__main__":
```
ninamiolane, line 19: No need for these 2 lines

toponetx/__init__.py
```diff
@@ -18,13 +18,17 @@
 from .classes.reportviews import CellView, HyperEdgeView, SimplexView, NodeView
 
 
-from .datasets.mesh import stanford_bunny
+from .datasets.mesh import stanford_bunny, shrec_16
```
ninamiolane, line 21: These imports are not in alphabetic order, I wonder why flake8 did not see this...

ninamiolane: --> They should be in alphabetic order

toponetx/algorithms/spectrum.py
```diff
@@ -117,8 +117,8 @@ def set_hodge_laplacian_eigenvector_attrs(
     Examples
     --------
     >>> from toponetx import SimplicialComplex
-    >>> SC=SimplicialComplex([[1,2,3],[2,3,5],[0,1]])
-    >>> SC = set_hodge_laplacian_eigenvector_attrs(SC,1,2,"down")
+    >>> SC=SimplicialComplex([[1, 2, 3],[2, 3, 5],[0, 1]])
```
ninamiolane, line 120:
```suggestion
    >>> SC = SimplicialComplex([[1, 2, 3], [2, 3, 5], [0, 1]])
```

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,92 @@
...
+)
+from toponetx.transform.graph_to_simplicial_complex import graph_2_clique_complex
+
+__all__ = ["karate_club"]
+
+
+def karate_club(complex_type="simplicial", feat_dim=2):
+    """Load the karate club as complex.
+
+        simplicial complex is returned as the clique complex of the graph
```
ninamiolane, line 19: Sentence needs to start with an uppercase, e.g. A simplicial complex ...
Ditto for the sentence below.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,92 @@
...
+        cell complex is return as the cell complex obtained by adding the independent cycles of graph
```
ninamiolane, line 20: typo: returned

ninamiolane: missing period at the end of the above sentences.

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Parameters
+    ----------
+    complex_type : str, optional
+        The type of complex to load. Supported values are
+        "simplicial complex" and "cell complex".
```
ninamiolane, line 26:
```suggestion
        "simplicial" and "cell".
```

toponetx/datasets/graph.py
```diff
@@ -0,0 +1,92 @@
...
+        The default is "simplicial complex".
```
ninamiolane, line 27:
```suggestion
        The default is "simplicial".
```

toponetx/datasets/graph.py
ninamiolane: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/no-else-return.html

ninamiolane (changes requested):

toponetx/__init__.py
josefhoppe: I think flake8 doesn't check that, but isort. And for some reason, isort ignores all \_\_init__.pys:

(from pyproject.toml)
```toml
[tool.isort]
multi_line_output = 3
include_trailing_comma = true
skip = [".gitignore", "__init__.py", ".venv"]
```

toponetx/datasets/graph.py
ffl096: In general the correct way would be to note the possible values as literal types, not as part of the parameter description. E.g.:
```
complex_type : {'simplicial', 'cell'}, default='simplicial'
    The type of complex to load.
```
In addition, `default='simplex'` is more appropriate than `optional`.

test/datasets/test_graph.py
```diff
@@ -0,0 +1,23 @@
...
+
+from toponetx import CellComplex, SimplicialComplex
+from toponetx.datasets.graph import karate_club
+
+
+class TestGraph:
+    """Test datasets utils."""
+
+    def test_karate_club(self):
+        """Test stanford_bunny."""
```
josefhoppe, line 11:
```suggestion
        """Test karate_club."""
```

toponetx/datasets/graph.py
```diff
@@ -13,15 +15,15 @@
 __all__ = ["karate_club"]
 
 
-def karate_club(complex_type="simplicial", feat_dim=2):
+def karate_club(
+    complex_type: Optional[str] = "simplicial", feat_dim: Optional[str] = 2
```
ffl096, line 19: - `complex_type` is not optional. The correct typing annotation would be `Literal["simplicial", "cell"]`
- `feat_dim` is not optional and is not of type `str`. The correct typing annotation would be just `int`.

Remember: `Optional` just means that `None` is an accepted value. This is not the case here. It has nothing to do with whether the caller has to provide a value or not. That is entirely determined by whether the parameter has a default value or not.

mhajij: great, thanks for being patient

mhajij: in that case, I do not need to check for the correct string when using Literal["simplicial", "cell"]?

toponetx/datasets/mesh.py
```diff
@@ -92,37 +91,37 @@ def shrec_16(size="full"):
     url = "https://github.com/mhajij/shrec_16/raw/main/shrec.zip"
     url_small = "https://github.com/mhajij/shrec_16/raw/main/small_shrec.zip"
     if size == "full":
```
josefhoppe, line 93: We have some duplicated code here, which adds unnecessary complexity. I'd suggest grouping the differences into one if/elif and then only having the download / extract logic once:
```python
    if size == "full":
        url = url
        zipfile = os.path.join(DIR, "shrec.zip")
        training = os.path.join(DIR, "shrec_training.npz")
        testing = os.path.join(DIR, "shrec_testing.npz")
    elif size == "small":
        url = url_small
        zipfile = os.path.join(DIR, "small_shrec.zip")
        training = os.path.join(DIR, "small_shrec_training.npz")
        testing = os.path.join(DIR, "small_shrec_testing.npz")
    else:
        raise ValueError(f"size must be 'full' or 'small' got {size}.")
    if not os.path.isfile(zipfile):
    ...
```

Could be further reduced to something like:
```python
    DS_MAP = {
        'full': ("shrec", "https://github.com/mhajij/shrec_16/raw/main/shrec.zip"),
        'small': ("small_shrec", "https://github.com/mhajij/shrec_16/raw/main/small_shrec.zip")
    }
    if size not in DS_MAP:
        raise ValueError(f"size must be 'full' or 'small' got {size}.")
    url, ds_name = DS_MAP[size]
    
    zipfile = os.path.join(DIR, f"{ds_name}.zip")
    training = os.path.join(DIR, f"{ds_name}_training.npz")
    testing = os.path.join(DIR, f"{ds_name}_testing.npz")
```

josefhoppe (changes requested):

toponetx/datasets/mesh.py
josefhoppe: I think the second option is better since it is shorter and shows the naming logic on the first look.

mhajij: thanks, yes it is much better

toponetx/datasets/graph.py
ffl096: You still should check for the value inside the function, it's not a big overhead. Type annotations are not enforced by Python itself.

josefhoppe (approved):

ffl096 (approved):

toponetx/datasets/mesh.py
```diff
@@ -89,40 +88,39 @@ def shrec_16(size="full"):
...
+
+    if not os.path.isfile(training):
+        print("downloading dataset...\n")
+        wget.download(url, DIR)
+        with zipfile.ZipFile(zip_file, "r") as zip_ref:
+            zip_ref.extractall(DIR)
+    if not os.path.isfile(testing):
+        wget.download(url, DIR)
+        with zipfile.ZipFile(zip_file, "r") as zip_ref:
+            zip_ref.extractall(DIR)
```
josefhoppe, line 114: Since both ifs do the same, we can merge them for simplicity.

```python
if not os.path.isfile(training) or not os.path.isfile(testing):
    ...
```

toponetx/datasets/mesh.py
```diff
@@ -89,40 +88,39 @@ def shrec_16(size="full"):
...
+    print("done!")
```
josefhoppe, line 115: Does this line belong to the download? Then it should be indented one level more to fit into the if above.

josefhoppe (changes requested):

ninamiolane (changes requested):
The functions loading datasets in the datasets/ folder should return datasets in the same format, i.e. complexes with features attached on their cells.

Currently, they return many different things:
- complex
- dictionary
- “tuple of length 2 npz files”
--> this needs to be fixed, in order to be consistent, which will help users moving from one dataset to another without having to worry about which format that dataset is going to be output as.

toponetx/datasets/mesh.py
ninamiolane: Fixed.

toponetx/datasets/mesh.py
ninamiolane: Fixed.

ninamiolane:
Also: `feat_dim: int = 2):`
“feat_dim” is not a feature dimension, it’s a number of eigenvectors n_eigenvectors. If each of the eigenvectors have dimension dim  then the actual feature dimension is n_eigenvectors * dim  : correct?
--> should it be renamed?

josefhoppe (approved):

mathildepapillon (approved):

mathildepapillon (approved):

mhajij:
> Also: `feat_dim: int = 2):` “feat_dim” is not a feature dimension, it’s a number of eigenvectors n_eigenvectors. If each of the eigenvectors have dimension dim then the actual feature dimension is n_eigenvectors * dim : correct? --> should it be renamed?

We discussed this one, essentially feat_dim here is the same as n_eigenvectors.

PR #137, opened 2023-06-02
Remove isort ignore of __init__.py files

josefhoppe:


toponetx/__init__.py
```diff
@@ -1,32 +1,26 @@
...
-from .classes.complex import Complex
-from .classes.simplicial_complex import SimplicialComplex
+from .classes.cell import Cell
 from .classes.cell_complex import CellComplex
 from .classes.combinatorial_complex import CombinatorialComplex
-
-from .classes.simplex import Simplex
+from .classes.complex import Complex
 from .classes.hyperedge import HyperEdge
-from .classes.cell import Cell
```
josefhoppe, line 16: It seems these imports are not sorted alphabetically because they are grouped by function. Someone must have made this decision, and I can see that it may actually be easier to find imports this way.

Overall, I'm not sure if we should go ahead with this PR or keep it the way it was.

josefhoppe: Since only one line is shown in the preview, here is a larger excerpt from the code

```python
from .classes.complex import Complex
from .classes.simplicial_complex import SimplicialComplex
from .classes.cell_complex import CellComplex
from .classes.combinatorial_complex import CombinatorialComplex

from .classes.simplex import Simplex
from .classes.hyperedge import HyperEdge
from .classes.cell import Cell
```

ninamiolane (approved):
Thanks! I think it's better to keep them alphabetically indeed. The folders already group code by function.

PR #138, opened 2023-06-02
Cleanup functions regarding dimension of complexes

ffl096:
All complexes have a `dim()` method as per the `Complex` contract. However, simplicial and cell complexes had an identical second method under another name for the same purpose.

This pull request deprecates these methods and unifies everything to the method dictated by the `Complex` contract for consistency. I kept the old methods with a warning for now since they may be used by some downstream consumers already -- lets remove them with a future release.


ninamiolane (approved):
Excellent, thank you!

PR #149, opened 2023-06-11
test cases for SimplicialComplex

gurug-dev:
few fixes and additional test coverage for SimplicialComplex

### DIFF
--- a/test/classes/test_simplicial_complex.py
+++ b/test/classes/test_simplicial_complex.py
@@ -334,6 +334,16 @@ class TestSimplicialComplex(unittest.TestCase):
             np.array([[0, 1, -1, 1, 0, 0], [0, 0, 0, 1, -1, 1]]).T,
         )
 
+        # repeat the same test, but with signed=False
+        B2 = SC.incidence_matrix(rank=2, signed=False)
+        assert B2.shape == (6, 2)
+
+        # assert that the incidence matrix is correct
+        np.testing.assert_array_equal(
+            B2.toarray(),
+            np.array([[0, 1, 1, 1, 0, 0], [0, 0, 0, 1, 1, 1]]).T,
+        )
+
     def test_hodge_laplacian_matrix(self):
         """Test hodge_laplacian_matrix shape and values."""
         # create a SimplicialComplex object with a few simplices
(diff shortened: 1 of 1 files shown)


test/classes/test_simplicial_complex.py
```diff
@@ -167,7 +168,19 @@ def test_setitem__(self):
         SC.add_simplex(s)
         assert SC["A"]["heat"] == 1
 
-    def add_simplices_from(self):
+    def test_maxdim(self):
+        """Test deprecated maxdim property for deprecation warning."""
+        with warnings.catch_warnings(record=True) as w:
```
ffl096, line 173: Please use [`pytest.deprecated_call`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.deprecated_call)

ffl096 (changes requested):

gurug-dev:
@ffl096 @mhajij fixed requested changes, modified deprecation testcase.

test/classes/test_simplicial_complex.py
```diff
@@ -321,6 +329,16 @@ def test_incidence_matrix_1(self):
...
+
+        # assert that the incidence matrix is correct
+        np.testing.assert_array_equal(
+            B2.toarray(),
+            np.array([[0, 1, 1, 1, 0, 0], [0, 0, 0, 1, 1, 1]]).T,
+        )
+
     def test_hodge_laplacian_matrix(self):
         """Test hodge_laplacian_matrix shape and values."""
         # create a SimplicialComplex object with a few simplices
```
mhajij, line 344: remove comments from code, they are hard to maintain

test/classes/test_simplicial_complex.py
```diff
@@ -344,6 +362,60 @@ def test_hodge_laplacian_matrix(self):
...
+                [0.0, 0.0, 1.0, 1.0, 0.0],
+            ]
+        )
+
+        np.testing.assert_array_equal(L_hodge.toarray(), np.abs(D - A))
+        expected_row = {(0,): 0, (1,): 1, (2,): 2, (3,): 3, (4,): 4}
+
+        self.assertDictEqual(row, expected_row)
+
+        # compute Hodge Laplacian for different ranks
```
mhajij, line 386: same, remove comments.

test/classes/test_simplicial_complex.py
```diff
@@ -167,7 +168,14 @@ def test_setitem__(self):
         SC.add_simplex(s)
         assert SC["A"]["heat"] == 1
 
-    def add_simplices_from(self):
+    def test_maxdim(self):
+        """Test deprecated maxdim property for deprecation warning."""
+        with pytest.deprecated_call():
+            # Cause a warning by accessing the deprecated maxdim property
+            SC = SimplicialComplex()
+            _ = SC.maxdim
```
ffl096, line 176: `SimplicialComplex.maxdim` should not only raise a deprecation warning but still also return the correct value. Maybe something like this would be better?

```python
with pytest.deprecated_call():
    SC = SimplicialComplex([(1, 2, 3)])
    assert SC.maxdim == 2
```

ffl096 (changes requested):

test/classes/test_simplicial_complex.py
gurug-dev: Missed to add this back! Will do

ffl096:
Failing tests are from fetching datasets remotely, I think we are making too many requests

PR #155, opened 2023-06-12
Sorting when converting set to tuple

### LINKED ISSUE #154: Mix of ordered and unordered data structures
The following example leads to an error:

```
from toponetx import SimplicialComplex as sc

simplices = [[3668, 5308, 5900]]
SC = sc(simplices)
B1 = SC.incidence_matrix(2)
```

The problem if I'm not mistaken is a mix of ordered data structures (tuple) and unordered (fronzen set):
- SimplicialComplex._simplex_set contains simplices as sets
- But in SimplicialComplex.skeleton function the simplices are converted to tuples
- In SimplicialComplex.incidence_matrix the faces are taken by set difference and converted to tuple, but this may not match the keys of the structure returned by SimplicialComplex.skeleton

Hellsegga:
Fixes #154

I'm not very familiar with the code base yet and not sure this is a good way to solve it.

### DIFF
--- a/toponetx/classes/simplicial_complex.py
+++ b/toponetx/classes/simplicial_complex.py
@@ -198,7 +198,7 @@ class SimplicialComplex(Complex):
         Set of simplices of dimension n.
         """
         if rank < len(self._simplex_set.faces_dict) and rank >= 0:
-            return sorted(tuple(i) for i in self._simplex_set.faces_dict[rank].keys())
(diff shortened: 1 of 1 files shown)


mhajij (changes requested):
Can you make sure the lint test pass? 

Make sure you do 

pre-commit install

then the lint changes will be done to you automatically.

ffl096 (changes requested):
Can you please add a test case (your example from #154 should be fine) to avoid any future regression?

mhajij (approved):

PR #158, opened 2023-06-13
Inline internal functions that are only called once

ffl096:
There is little reason why we should have e.g. a public `add_simplex` function that only forwards `_add_simplex`.


mhajij (approved):

mhajij:
@ffl096  There are conflict with the main branch--can you please double check?

ffl096:
(failing tests are again unrelated)

PR #163, opened 2023-06-14
Add static typechecking with mypy

### LINKED ISSUE #102: Add hints for parameters and return values!

ffl096:
This is a draft pull request to track the progress of adding typing annotations to our codebase. For now it is expected that mypy raises errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


ninamiolane:
@ffl096 update on this? Should we close?

ffl096:
There are still many errors in the code base. This pull request is meant to track the progress of that.

ffl096:
Let's get this merged as is for now. There are many typing issues remaining, but most of them are non-trivial to fix. We can keep an eye on it via the GitHub Action (that is allowed to fail) but don't enforce compliance with mypy on TopoNetX for now.

mhajij (approved):

PR #165, opened 2023-06-14
Remove uses of `unittest` in favour of `pytest`

### LINKED ISSUE #160: Future test framework
As of right now, the test suite is a mixture of `unittest` and `pytest`, essentially bounding the tests to both modules:
- Some tests inherent from `unittest.TestCase` and make use of `self.assert*` functions. While `pytest` is happy to run these tests for the moment, some `unittest` functions like [subtests to distinguish iterations](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) are not supported.
- Other tests use `assert` statements or are explicitly bound to `pytest` (e.g., using `pytest.deprecated_call()`).

Let's decide on one of both frameworks and refactor all tests to no longer use the other framework. This should also be enforced in new pull requests.

ffl096:
closes #160

### DIFF
--- a/test/algorithms/test_spectrum.py
+++ b/test/algorithms/test_spectrum.py
@@ -1,7 +1,6 @@
 """Test spectrum module."""
 
-import unittest
-
+import pytest
 import scipy.sparse as sparse
 
 from toponetx.algorithms.spectrum import (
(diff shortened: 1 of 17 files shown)


ninamiolane (approved):
Yesss! 🎉 This was long overdue, many thanks!

Issue #161, opened 2023-06-13
Minimum Python Version


ffl096:
Minimum Python versions are kind of a mess across all packages:
- TopoNetX is Python 3.9+
- TopoEmbedX and TopoModelX are Python 3.10+

Moreover, the Python version in the README files is wrong:
- Even though TopoNetX is 3.9+, the README claims 3.10 is required.
- Even though TopoModelX is 3.10+, the EADME claims 3.11 is required.

Lets decide on a minimum version that applies to all three packages and fix the README files accordingly.

ffl096:
I would say to at least support Python 3.10. Dropping 3.9 comes mostly down to some typing features that are not strictly necessary but are nice-to-have.

ninamiolane:
Agreed, let's do 3.10

PR #168, opened 2023-06-15
Unify python version across packages

ffl096:
As outlined in #161, the minimum Python version is inconsistent across our packages, which will be fixed by this pull request and others for the other repositories.

For `TopoNetX`, this moves the package from Python 3.9+ to Python 3.10+, allowing us to use some new typing features. In particular, `typing.Optional` and `typing.Union` are deprecated and replaced by [union type expressions](https://docs.python.org/3/library/stdtypes.html#types-union).


ffl096:
Tests should pass once #165 is merged

PR #176, opened 2023-06-25
normalization added

mhajij:
often in DL computations normalization must be added to matrices for computations. This file is a collection of popular methods typically used in a DL setting.

### DIFF
--- a/test/utils/test_normalization.py
+++ b/test/utils/test_normalization.py
@@ -25,12 +25,12 @@ def test_compute_laplacian_normalized_matrix():
     """Test normalize laplacian."""
     adjacency_matrix = np.array(
         [
-            [0, 1, 0, 0, 0, 1],
-            [1, 0, 1, 0, 0, 0],
-            [0, 1, 0, 1, 0, 0],
-            [0, 0, 1, 0, 1, 0],
-            [0, 0, 0, 1, 0, 1],
-            [1, 0, 0, 0, 1, 0],
+            [0.0, 1.0, 0.0, 0.0, 0.0, 1.0],
+            [1.0, 0.0, 1.0, 0.0, 0.0, 0.0],
+            [0.0, 1.0, 0.0, 1.0, 0.0, 0.0],
+            [0.0, 0.0, 1.0, 0.0, 1.0, 0.0],
+            [0.0, 0.0, 0.0, 1.0, 0.0, 1.0],
+            [1.0, 0.0, 0.0, 0.0, 1.0, 0.0],
         ]
     )
 
@@ -42,7 +42,7 @@ def test_compute_laplacian_normalized_matrix():
 
     L = csr_matrix(L).asfptype()
     normalized_L = compute_laplacian_normalized_matrix(L)
-    expected_result = csr_matrix(
+    expected_result = np.array(
         [
             [0.5, -0.25, 0.0, 0.0, 0.0, -0.25],
             [-0.25, 0.5, -0.25, 0.0, 0.0, 0.0],
@@ -67,7 +67,7 @@ def test_compute_x_laplacian_normalized_matrix():
     L = csr_matrix([[4.0, 0], [0.0, 4.0]])
     Lx = csr_matrix([[0.0, 1.0], [1.0, 0.0]])
     normalized_Lx = compute_x_laplacian_normalized_matrix(L, Lx)
-    expected_result = csr_matrix([[0.0, 0.25], [0.25, 0.0]])
+    expected_result = np.array([[0.0, 0.25], [0.25, 0.0]])
     assert np.allclose(normalized_Lx.toarray(), expected_result.toarray())
 
 
@@ -85,7 +85,7 @@ def test_compute_kipf_adjacency_normalized_matrix():
         ]
     )
     normalized_A_opt = compute_kipf_adjacency_normalized_matrix(csr_matrix(A_opt))
-    expected_result = csr_matrix(
+    expected_result = np.array(
         [
             [0.0, 0.5, 0.0, 0.0, 0.0, 0.5],
             [0.5, 0.0, 0.5, 0.0, 0.0, 0.0],
@@ -101,7 +101,7 @@ def test_compute_kipf_adjacency_normalized_matrix():
     normalized_A_opt = compute_kipf_adjacency_normalized_matrix(
         csr_matrix(A_opt), add_identity=True
     )
-    expected_result = csr_matrix(
+    expected_result = np.array(
         [
             [0.33333333, 0.33333333, 0.0, 0.0, 0.0, 0.33333333],
             [0.33333333, 0.33333333, 0.33333333, 0.0, 0.0, 0.0],
--- a/test/utils/test_normalization.py
+++ b/test/utils/test_normalization.py
@@ -1,0 +1,117 @@
+"""Test normalization."""
+
+import numpy as np
+import scipy.sparse.linalg as spl
+from scipy.sparse import coo_matrix, csr_matrix, diags
+
+from toponetx.utils.normalization import (
+    asymmetric_kipf_normalization,
+    asymmetric_xu_normalization,
+    kipf_adjacency_matrix_normalization,
+    normalize_laplacian,
+    normalize_x_laplacian,
+)
+
+
+def test_normalize_laplacian():
+    """Test normalize laplacian."""
+    adjacency_matrix = np.array(
+        [
+            [0, 1, 0, 0, 0, 1],
+            [1, 0, 1, 0, 0, 0],
+            [0, 1, 0, 1, 0, 0],
+            [0, 0, 1, 0, 1, 0],
+            [0, 0, 0, 1, 0, 1],
+            [1, 0, 0, 0, 1, 0],
+        ]
+    )
+
+    # Calculate the degree matrix
+    degree_matrix = np.diag(adjacency_matrix.sum(axis=1))
+
+    # Calculate the Laplacian matrix
+    L = np.array(degree_matrix - adjacency_matrix)
+
+    L = csr_matrix(L).asfptype()
+    normalized_L = normalize_laplacian(L)
+    expected_result = csr_matrix(
+        [
+            [0.5, -0.25, 0.0, 0.0, 0.0, -0.25],
+            [-0.25, 0.5, -0.25, 0.0, 0.0, 0.0],
+            [0.0, -0.25, 0.5, -0.25, 0.0, 0.0],
+            [0.0, 0.0, -0.25, 0.5, -0.25, 0.0],
+            [0.0, 0.0, 0.0, -0.25, 0.5, -0.25],
+            [-0.25, 0.0, 0.0, 0.0, -0.25, 0.5],
+        ]
+    )
+    assert np.allclose(normalized_L.toarray(), expected_result.toarray())
+
+
+def test_normalize_x_laplacian():
+    """Test normalize up or down laplacian."""
+    L = csr_matrix([[2.0, -1.0, 0.0], [-1.0, 3.0, -1.0], [0.0, -1.0, 2.0]])
+    Lx = csr_matrix([[1.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 1.0, 0.0]])
+    normalized_Lx = normalize_x_laplacian(L, Lx)
+    expected_result = csr_matrix([[0.25, 0.0, 0.0], [0.0, 0.0, 0.25], [0.0, 0.25, 0.0]])
+    assert np.allclose(normalized_Lx.toarray(), expected_result.toarray())
+
+    # Test case 2
+    L = csr_matrix([[4.0, 0], [0.0, 4.0]])
+    Lx = csr_matrix([[0.0, 1.0], [1.0, 0.0]])
+    normalized_Lx = normalize_x_laplacian(L, Lx)
+    expected_result = csr_matrix([[0.0, 0.25], [0.25, 0.0]])
+    assert np.allclose(normalized_Lx.toarray(), expected_result.toarray())
+
+
+def test_kipf_adjacency_matrix_normalization():
+    """Test kipf_adjacency_matrix_normalization."""
+    # Test case 1
+    A_opt = np.array([[0, 1, 0], [1, 0, 1], [0, 1, 0]])
+    normalized_A_opt = kipf_adjacency_matrix_normalization(A_opt)
+    expected_result = coo_matrix(
+        [[0.0, 0.70710678, 0.0], [0.70710678, 0.0, 0.70710678], [0.0, 0.70710678, 0.0]]
+    )
+    assert np.allclose(normalized_A_opt.toarray(), expected_result.toarray())
+
+    # Test case 2
+    A_opt = np.array([[1, 1], [1, 0]])
+    normalized_A_opt = kipf_adjacency_matrix_normalization(A_opt)
+    expected_result = coo_matrix([[0.70710678, 0.70710678], [0.89442719, 0.0]])
+    assert np.allclose(normalized_A_opt.toarray(), expected_result.toarray())
+
+
+def test_asymmetric_kipf_normalization():
+    """Unit test for asymmetric_kipf_normalization function."""
+    # Test case 1: Sparse matrix input
+    A = np.array([[0, 1, 1], [1, 0, 0], [0, 1, 0]])
+    expected_result = coo_matrix(
+        [[0.0, 0.70710678, 0.70710678], [0.70710678, 0.0, 0.0], [0.0, 1.0, 0.0]]
+    )
+    normalized_matrix = asymmetric_kipf_normalization(A, is_sparse=True)
+    assert np.allclose(normalized_matrix.toarray(), expected_result.toarray())
+
+    # Test case 2: Dense matrix input
+    B = np.array([[0, 2, 1], [1, 0, 0], [0, 1, 0]])
+    expected_result = np.array(
+        [[0.0, 0.89442719, 0.4472136], [0.4472136, 0.0, 0.0], [0.0, 0.89442719, 0.0]]
+    )
+    normalized_matrix = asymmetric_kipf_normalization(B, is_sparse=False)
+    assert np.allclose(normalized_matrix, expected_result)
+
+    # Test case 3: Empty matrix
+    C = np.array([])
+    expected_result = np.array([])
+    normalized_matrix = asymmetric_kipf_normalization(C)
+    assert np.allclose(normalized_matrix, expected_result)
+
+    # Test case 4: Single-row matrix
+    D = np.array([[0, 1, 0]])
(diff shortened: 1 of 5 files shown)


toponetx/utils/normalization.py
```diff
@@ -0,0 +1,359 @@
...
+        L (scipy.sparse.csr_matrix): The Laplacian matrix.
+
+    Returns
+    -------
+        scipy.sparse.csr_matrix: The normalized Laplacian matrix.
+
+    Notes
+    -----
+        This function normalizes the Laplacian matrix by dividing it by the largest eigenvalue.
+    """
```
ffl096, line 40: The documentation style in this file does not follow NumPy's Conventions and will make problems with the documentation website. I'll illustrate how it should look like based on this example, can you please update it throughout the file?

```python3
"""Normalise the Laplacian matrix.

This function normalises the Laplacian matrix by dividing it by its largest eigenvalue.

Parameters
----------
L : scipy.sparse.scr_matrix
    The Laplacian matrix.

Returns
-------
scipy.sparse.csr_matrix
    The normalised Laplacian matrix.
"""

File: toponetx/utils/normalization.py [hunk 2]
```diff
@@ -0,0 +1,359 @@
...
+    topeigen_val = spl.eigsh(L.asfptype(), k=1, which="LM", return_eigenvectors=False)[
+        0
+    ]
+    out = L.copy()
+    out *= 1.0 / topeigen_val
+    return out
```
ffl096, line 46: This looks more complicated than it needs to be, no? You don't need to copy explicitly.
```python3
return L / topeigen_val
```

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,359 @@
...
+    Notes
+    -----
+        This function normalizes the up or down Laplacian matrices by dividing it
+        by the largest eigenvalue of the Laplacian matrix.
+    """
+    assert L.shape[0] == L.shape[1]
+    topeig = spl.eigsh(L.asfptype(), k=1, which="LM", return_eigenvectors=False)[0]
+    out = Lx.copy()
+    out *= 1.0 / topeig
+    return out
```
ffl096, line 70: See above, can be simplified.

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,359 @@
...
+    Parameters
+    ----------
+    B1 : numpy array or scipy coo_matrix
+        The boundary B1: C1->C0 of a complex.
+    B2 : numpy array or scipy coo_matrix
+        The boundary B2: C2->C1 of a complex.
+
+    Returns
+    -------
+    _ : numpy array or scipy coo_matrix
```
ffl096, line 197:
```suggestion
np.ndarray or scipy.sparse.coo_matrix
```

ffl096 (changes requested):

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,359 @@
...
+        by the largest eigenvalue of the Laplacian matrix.
+    """
+    assert L.shape[0] == L.shape[1]
+    topeig = spl.eigsh(L.asfptype(), k=1, which="LM", return_eigenvectors=False)[0]
+    out = Lx.copy()
+    out *= 1.0 / topeig
+    return out
+
+
+def kipf_adjacency_matrix_normalization(
```
michaelschaub, line 73: I think we should use descriptive names for these functions (in that the names should indicate what the function does).

Something like:
normalized_symmetric_adjacency_matrix

is what is computed here. It is true (and perhaps) interesting that this is used in Kipf's work, but it is not an invention of that paper as well..
The information where this normalization is used could be provided in the function description with a reference. Examples: Kipf... [REF]

Second, it would be good to have all names consistent, atm we have things like

normalize_x_laplacian (verb first), and ..._adjacency_matrix_normalization.

We should decide for one style and apply consistently, Perhaps something like:

compute_xy_normalized_zmatrix ??

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,359 @@
...
+        return normalized_matrix
+    else:
+        Di = np.sum(np.abs(B), axis=1)
+        Di2 = np.diag(np.power(Di, -1))
+        Di2[np.isinf(Di2)] = 0.0
+        B = np.array(B)
+        normalized_matrix = Di2.dot
+
+
+def bunch_normalization(B1, B2):
```
michaelschaub, line 145: see above re: naming..

toponetx/utils/normalization.py
mhajij: Sure, one question, we should be consistent with the sparse type all over. Which type should we adapt? csr_matrix or coo?

michaelschaub: The output should be csr -- coo is really only for construction of matrices
See intended usage here
https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html

toponetx/utils/normalization.py
ninamiolane: I agree.

mhajij: @michaelschaub  do you know a better reference for the above normalization ?

Also, do you have a ref for D^{-1}A normalization? 

please let me know if you like the current naming convection-I can change it again if you have something better in mind.

toponetx/__init__.py
```diff
@@ -21,6 +21,21 @@
...
 )
+from .utils.normalization import (
+    _compute_B1_normalized_matrix,
+    _compute_B1T_normalized_matrix,
+    _compute_B2_normalized_matrix,
+    _compute_B2T_normalized_matrix,
+    _compute_D1,
+    _compute_D2,
+    _compute_D3,
+    _compute_D5,
```
ffl096, line 32: These are internal functions (starting with an underscore) and should not be imported into the `toponetx` namespace.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+
+from toponetx.utils.normalization import (
+    _compute_B1_normalized_matrix,
+    _compute_B1T_normalized_matrix,
+    _compute_B2_normalized_matrix,
+    _compute_B2T_normalized_matrix,
+    _compute_D1,
+    _compute_D2,
+    _compute_D3,
+    _compute_D5,
```
ffl096, line 15: As far as I see these internal functions are not called directly here and should not be imported.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Calculate the degree matrix
+    degree_matrix = np.diag(adjacency_matrix.sum(axis=1))
+
+    # Calculate the Laplacian matrix
+    L = np.array(degree_matrix - adjacency_matrix)
+
+    L = csr_matrix(L).asfptype()
+    normalized_L = compute_laplacian_normalized_matrix(L)
+    expected_result = csr_matrix(
```
ffl096, line 45: No need to construct a `csr_matrix` here when you convert it to dense arrays below again. Just use `np.array`.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    assert np.allclose(normalized_L.toarray(), expected_result.toarray())
+
+
+def test_compute_x_laplacian_normalized_matrix():
+    """Test normalize up or down laplacian."""
+    L = csr_matrix([[2.0, -1.0, 0.0], [-1.0, 3.0, -1.0], [0.0, -1.0, 2.0]])
+    Lx = csr_matrix([[1.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 1.0, 0.0]])
+    normalized_Lx = compute_x_laplacian_normalized_matrix(L, Lx)
+    expected_result = csr_matrix([[0.25, 0.0, 0.0], [0.0, 0.0, 0.25], [0.0, 0.25, 0.0]])
```
ffl096, line 63: See above.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+    assert np.allclose(normalized_Lx.toarray(), expected_result.toarray())
+
+    # Test case 2
+    L = csr_matrix([[4.0, 0], [0.0, 4.0]])
+    Lx = csr_matrix([[0.0, 1.0], [1.0, 0.0]])
+    normalized_Lx = compute_x_laplacian_normalized_matrix(L, Lx)
+    expected_result = csr_matrix([[0.0, 0.25], [0.25, 0.0]])
```
ffl096, line 70: See above.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+            [0, 1, 0, 0, 0, 1],
+            [1, 0, 1, 0, 0, 0],
+            [0, 1, 0, 1, 0, 0],
+            [0, 0, 1, 0, 1, 0],
+            [0, 0, 0, 1, 0, 1],
+            [1, 0, 0, 0, 1, 0],
+        ]
+    )
+    normalized_A_opt = compute_kipf_adjacency_normalized_matrix(csr_matrix(A_opt))
+    expected_result = csr_matrix(
```
ffl096, line 88: See above.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,183 @@
...
+            [0.5, 0.0, 0.0, 0.0, 0.5, 0.0],
+        ]
+    )
+
+    assert np.allclose(normalized_A_opt.toarray(), expected_result.toarray())
+
+    normalized_A_opt = compute_kipf_adjacency_normalized_matrix(
+        csr_matrix(A_opt), add_identity=True
+    )
+    expected_result = csr_matrix(
```
ffl096, line 104: See above.

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+    "compute_xu_asymmetric_normalized_matrix",
+    "compute_bunch_normalized_matrices",
+    "_compute_B1_normalized_matrix",
+    "_compute_B1T_normalized_matrix",
+    "_compute_B2_normalized_matrix",
+    "_compute_B2T_normalized_matrix",
+    "_compute_D1",
+    "_compute_D2",
+    "_compute_D3",
+    "_compute_D5",
```
ffl096, line 22: Remove the internal functions (starting with an underscore) from `__all__`, please :)

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    Normalize the adjacency matrix using Kipf's normalization.
+
+    Typically used to normalize adjacency matrices.
+
+    Parameters
+    ----------
+    A_opt : numpy.ndarray
+        The adjacency matrix.
+    add_identity : bool, optional
```
ffl096, line 88:
```suggestion
    add_identity : bool, default=False
```
In general optional boolean flags feel weird, there is nothing optional about them. They default to true or to false, that's it.

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+
+    Returns
+    -------
+    scipy.sparse.csr_matrix
+        The normalized adjacency matrix.
+
+    Notes
+    -----
+    This normalization is based on Kipf's formulation, which computes the row-sums,
+    constructs a diagonal matrix D^(-0.5), and applies the normalization as D^(-0.5) * A_opt * D^(-0.5).
```
ffl096, line 102: Maybe format this a proper LaTeX formulas? This would render better on the documentation website. See https://numpydoc.readthedocs.io/en/latest/format.html#notes and their rendered example.

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+    Compute Xu's normalized asymmetric matrix.
+
+    Typically used to normalize boundary operators.
+
+    Parameters
+    ----------
+    B : numpy.ndarray or scipy.sparse.csr_matrix
+        The asymmetric matrix.
+    is_sparse : bool, optional
+        If True, treat B as a sparse matrix.
```
ffl096, line 130: I wonder whether that parameter is actually useful? Just test whether `B` is sparse ([`scipy.sparse.issparse`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.issparse.html#scipy.sparse.issparse)) should be sufficient, right? Or is there ever a case where `B` is of sparse type but you want to set `is_sparse=False`?

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+        normalized_matrix = Di2.dot(B)
+        return normalized_matrix
+
+
+def compute_bunch_normalized_matrices(B1, B2):
+    """Get Bunch normalization.
+
+    Parameters
+    ----------
+    B1 : numpy array or scipy csr_matrix
```
ffl096, line 163: `numpy array` and `scipy csr_matrix` should be replaced by `np.ndarray` and `scipy.sparse.csr_matrix`, respectively, throughout the document

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,383 @@
...
+        The boundary B1: C1->C0 of a complex.
+    B2 : numpy array or scipy csr_matrix
+        The boundary B2: C2->C1 of a complex.
+
+    Returns
+    -------
+    _ : numpy array or scipy csr_matrix
```
ffl096, line 210: Please remove the `_ : ` here and below. If a function returns just one value (as opposed to a tuple of values), just state the type.

ffl096 (changes requested):

ninamiolane:
@mhajij @ffl096 update on this?

test/utils/test_normalization.py
```diff
@@ -67,7 +67,7 @@ def test_compute_x_laplacian_normalized_matrix():
     L = csr_matrix([[4.0, 0], [0.0, 4.0]])
     Lx = csr_matrix([[0.0, 1.0], [1.0, 0.0]])
     normalized_Lx = compute_x_laplacian_normalized_matrix(L, Lx)
-    expected_result = csr_matrix([[0.0, 0.25], [0.25, 0.0]])
+    expected_result = np.array([[0.0, 0.25], [0.25, 0.0]])
     assert np.allclose(normalized_Lx.toarray(), expected_result.toarray())
```
ffl096: The `toarray` call on `expected_result` is not necessary anymore after your changes.

toponetx/utils/normalization.py
```diff
@@ -1,5 +1,6 @@
-"""Normalize of adjacency, boundary matrices of complexes."""
+"""Normalize of Laplacians, (co)adjacency, boundary matrices of complexes."""
 
+from typing import Tuple, Union
```
ffl096, line 3: Please don't use `Tuple` and `Union` from the typing module:
- Instead of `typing.Tuple`, use the built-in `tuple`. It works exactly the same.
- Instead of `Union[A, B]` use the union type expression `A | B`.

toponetx/utils/normalization.py
```diff
@@ -74,26 +73,25 @@ def compute_x_laplacian_normalized_matrix(L, Lx):
...
+    A_opt: ndarray, add_identity: bool = False, identity_multiplier: float = 1.0
+) -> csr_matrix:
+    """Normalize the adjacency matrix using Kipf's normalization.
 
     Typically used to normalize adjacency matrices.
 
     Parameters
     ----------
-    A_opt : numpy.ndarray
+    A_opt : ndarray
```
ffl096, line 85: This should be reverted. Always use the full name in documentation (as per NumPy recommendation)

toponetx/utils/normalization.py
```diff
@@ -116,22 +114,23 @@ def compute_kipf_adjacency_normalized_matrix(
     return A_opt_normalized
 
 
-def compute_xu_asymmetric_normalized_matrix(B, is_sparse=True):
-    """
-    Compute Xu's normalized asymmetric matrix.
+def compute_xu_asymmetric_normalized_matrix(
+    B: Union[ndarray, csr_matrix], is_sparse: bool = True
```
ffl096, line 119: Use union type expression please:
```suggestion
    B: np.ndarray | csr_matrix, is_sparse: bool = True
```

toponetx/utils/normalization.py
```diff
@@ -1,5 +1,6 @@
-"""Normalize of adjacency, boundary matrices of complexes."""
+"""Normalize of Laplacians, (co)adjacency, boundary matrices of complexes."""
 
+from typing import Tuple, Union
 import numpy as np
 import scipy.sparse.linalg as spl
 from numpy import ndarray
```
ffl096, line 7: Please don't import `ndarray` like this. It is standard to import `numpy as np` and access everything through that import, i.e., `np.ndarray`.

ffl096 (changes requested):

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,397 @@
...
+    ----------
+    [1] Michael T. Schaub, Austin R. Benson,
+        Paul Horn, Gabor Lippner,
+        Ali Jadbabaie Random walks on
+        simplicial complexes and the normalized
+        hodge 1-laplacian.
+    [2] Eric Bunch, Qian You,
+        Glenn Fung, Vikas Singh
+        Simplicial 2-Complex Convolutional
+        Neural Networks.
```
devendragovil, line 195: I don't think this is the right way of introducing references. Please change.

test/utils/test_normalization.py
```diff
@@ -0,0 +1,168 @@
...
+            [0.0, 0.33333333, 0.0],
+            [0.33333333, 0.33333333, 0.33333333],
+        ],
+    )
+
+    assert np.allclose(
+        B2TN.toarray(),
+        np.array(
+            [[0.5, 0.0, 0.33333333], [0.0, 1.0, 0.33333333], [0.5, 0.0, 0.33333333]]
+        ),
```
devendragovil, line 167: There are a lot of distinct test cases that have been put together under one test function. Please separate them. Best practice is to have distinct test cases for different tests for easy isolation of issues, and debugging

devendragovil (changes requested):
Requesting some changes.

toponetx/utils/normalization.py
```diff
@@ -0,0 +1,393 @@
...
+        Normalized B2T: C1->C2.
+
+    References
+    ----------
+    [1] Schaub, M. T., Benson, A. R., Horn, P., Lippner, G., & Jadbabaie, A.
+        "Random walks on simplicial complexes and the normalized hodge 1-laplacian."
+
+    [2] Bunch, E., You, Q., Fung, G., & Singh, V.
+        "Simplicial 2-Complex Convolutional Neural Networks."
+    """
```
devendragovil, line 192: The references section is still not in the required format. See https://github.com/pyt-team/TopoModelX/blob/main/topomodelx/nn/simplicial/dist2cycle_layer.py for an example or here https://bashtage.github.io/sphinx-material/rst-cheatsheet/rst-cheatsheet.html

devendragovil (changes requested):
References need to be reformatted.

devendragovil (approved):
LGTM.

PR #180, opened 2023-06-27
Addition of coauthorship dataset

Hellsegga:
This is the dataset used in the Simplicial Neural Networks paper (https://arxiv.org/abs/2010.03633). I'd like to include it here to be able to use it in my implementation in TopoModelX (https://github.com/pyt-team/TopoModelX/pull/98). It can hopefully be useful to test other methods as well.


toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
         return cx
 
     raise ValueError(f"complex_type must be 'simplicial' or 'cell' got {complex_type}")
+
+
+def coauthorship() -> SimplicialComplex:
+    """Load the coauthorship network from [SNN20] as a simplicial complex.
```
ffl096, line 144: Empty line after this please.

toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
...
+    References
+    ----------
+    [SNN20] Stefania Ebli, Michael Defferrard and Gard Spreemann.
+        Simplicial Neural Networks.
+        Topological Data Analysis and Beyond workshop at NeurIPS.
+        https://arxiv.org/abs/2010.03633
+        https://github.com/stefaniaebli/simplicial_neural_networks
+
+    Returns
+    -------
```
ffl096, line 160: Please format the `Returns` section according to NumPy's documentation style:

```
Returns
-------
SimplicialComplex
    A simplicial complex obtained from the coauthorship ....
```

toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
...
+    A SimplicialComplex obtained from the coauthorship network in [SNN20].
+    The simplicial complex comes with the following feature:
+        "citations":
+            - the number of citations attributed to the given collaborations of k authors.
+
+    """
+    cochains = np.load(DIR / "150250_cochains.npy", allow_pickle=True)
```
ffl096, line 167: This file does not exist and I presume the user is expected to download the file themself? Then this should print a proper explanation of from where to obtain the necessary file.

toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
...
+
+    simplices = []
+    for dim in list(range(len(cochains)))[::-1]:
```
ffl096, line 170: No need to explicitly construct a list, use the `step` parameter of `range`:
```suggestion
    for dim in range(len(cochains) - 1, -1, -1):
```

toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
...
+        li = list(cochains[dim].keys())
+        simplices += [list(l) for l in li]
```
ffl096, line 172: No need to explicitly construct the list here, you are only iterating over its values below, which you can do with the iterator you have in the first place:
```suggestion
        simplices += [list(l) for l in cochains[dim].keys()]
```

toponetx/datasets/graph.py
```diff
@@ -135,3 +138,43 @@ def karate_club(
...
+
+    sc = SimplicialComplex(simplices)
+
+    for i in range(len(cochains)):
+        dic = {tuple(sorted(list(k))): v for k, v in cochains[i].items()}
```
ffl096, line 177: Again, explicit list construction is not necessary, `sorted` happily works on an iterable:
```suggestion
        dic = {tuple(sorted(k)): v for k, v in cochains[i].items()}
```

ffl096 (changes requested):

toponetx/datasets/graph.py
Hellsegga: The file is in the PR and I thought it could be part of the TopoNetX repo. Since it's a small file only available for download on the original repo of the author of the paper, maybe this is the best way? (same as for stanford bunny in mesh.py).

Hellsegga (commented):
Fixed in the last commit + see comment about storage of dataset.

toponetx/datasets/graph.py
ffl096: Sorry, my bad. I totally overlooked that the file is part of the pull request.

ffl096 (approved):

USFCA-MSDS (commented):
I think this has been addressed by now.

toponetx/datasets/graph.py
USFCA-MSDS, line 166: Can you change the name of the file?

toponetx/datasets/graph.py
```diff
@@ -9,7 +10,9 @@
 from toponetx.algorithms.spectrum import hodge_laplacian_eigenvectors
 from toponetx.transform.graph_to_simplicial_complex import graph_2_clique_complex
 
-__all__ = ["karate_club"]
+__all__ = ["karate_club", "coauthorship"]
```
USFCA-MSDS, line 13: I am not sure the addition of this dataset to graph is justified, maybe it needs a different file all together. Can you justify your choice for graph? other people might not find it intuitive anyway,

USFCA-MSDS (changes requested):

toponetx/datasets/graph.py
ninamiolane: Agreed, let's choose one name for this dataset and stick to it:
"coauthorship", then "coauthorship.npy" and the variable itself can be called `coauthorship`

toponetx/datasets/graph.py
Hellsegga: Yes, can you suggest a name for the file?
I'll be away for a week, I can fix the code based on the comments when I'm back.

Hellsegga:
I have fixed the changes suggested above.

However please advise about the file name. The dataset originates from a citation graph which is why I placed it in graph.py first, but it is already pre-processed as a simplicial complex and therefore I agree graph.py may not be ideal place. But I'm not sure where it should be placed.

ninamiolane:
@USFCA-MSDS we need to merge this to allow the challenge's participant @Hellsegga to move forward with their submission.
--> If there are remaining issues with the code, could you point them out so that @Hellsegga can create a follow-up PR? Thanks!

PR #182, opened 2023-06-28
Fixing little bug in CombinatorialComplex class

ManuelLecha:
_incidence_matrix method calling self.skeleton with an extra argument level, which is not needed nor defined in the CC class.
coadjacency_matrix method sending to incidence_to_adjacency method the incidence matrix transpose.


USFCA-MSDS (changes requested):
can you add unit tests for your changes to see what the issue you resolved?

ffl096 (approved):
Please rebase onto main so this can be merged 👍

USFCA-MSDS (approved):

PR #190, opened 2023-07-09
Adding test cases for remove nodes method for Cell Complex Class

devendragovil:
This commit was also used as a demonstration for Maneel


ffl096:
I would suggest the following to improve the quality of the test: Add a cell to the complex that contains additional nodes that are not to be deleted. Then assert that only these are present after the delete operation.
As of right now, a mistake in the delete method that just deletes everything would not be catched by the unittest.

mhajij (approved):

PR #206, opened 2023-07-22
Fixing adding cell complexity

mhajij:
The current implementation of adding cell in a combinatorial complex is expensive (creates a simplicial complex). Fixing the running time to make it O(n) where n is the size of the added cell


USFCA-MSDS (approved):

USFCA-MSDS:
Merge is needed to avoid further diverge. Of this commit from the unit tests.

PR #217, opened 2023-08-17
ColoredHyperGraph and CombinatorialComplex refactor 

gurug-dev:
NEW: Add ColoredHyperGraph base class 
REFACTOR: CombinatorialComplex to inherit CHG class


gurug-dev:
CombinatorialComplex class will be replaced with CombinatorialComplex2 after approving changes, codecov report is indicating 50% coverage due to the unused old class.

PR #220, opened 2023-08-18
First Implementation of a Simplex Trie

ffl096:
This implements a simplex trie as presented in [1] as backend data structure for the `SimplicialComplex` class. This is also used in gudhi's SC implementation. However, they do not expose all functionality we need and the data structure is implemented in native code, so we cannot interact with it directly either.

Using a simplex tree should bring some nice performance improvements over the previous approach and fixes some bugs along the way as well. I will add some comparisons later.

[1] Jean-Daniel Boissonnat and Clément Maria. The Simplex Tree: An Efficient Data Structure for General Simplicial Complexes. Algorithmica, pages 1–22, 2014

### DIFF
--- a/test/classes/test_combinatorial_complex.py
+++ b/test/classes/test_combinatorial_complex.py
@@ -477,6 +477,7 @@ class TestCombinatorialComplex:
                 frozenset({6}): {"weight": 1},
             }
         }
+
         example.remove_nodes(HyperEdge([3]))
         assert example._complex_set.hyperedge_dict == {
             0: {frozenset({4}): {"weight": 1}, frozenset({6}): {"weight": 1}}
--- a/test/classes/test_reportviews.py
+++ b/test/classes/test_reportviews.py
@@ -576,12 +576,10 @@ class TestReportViews_SimplexView:
 
     def test_getitem(self):
         """Test __getitem__ method of the SimplexView."""
-        assert self.simplex_view.__getitem__((1, 2)) == {
-            "is_maximal": True,
-            "membership": set(),
-        }
+        assert self.simplex_view[(1, 2)] == {}
+        assert self.simplex_view[1] == {}
         with pytest.raises(KeyError):
-            self.simplex_view.__getitem__([5])
+            _ = self.simplex_view[(5,)]
 
     def test_str(self):
         """Test __str__ method of the SimplexView."""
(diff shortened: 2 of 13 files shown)


ffl096:
@mhajij The tests fail because `coseg` loads a pickled state of `SimplicialComplex` with internal properties. This is (unrelated to this pull request) a bad idea, as any change of the data structure may lead to errors, or worse undetected inconsistencies.

mhajij:
> @mhajij The tests fail because `coseg` loads a pickled state of `SimplicialComplex` with internal properties. This is (unrelated to this pull request) a bad idea, as any change of the data structure may lead to errors, or worse undetected inconsistencies.

@ffl096 
I am not sure we should merge this pull request now because the ICML challenge participants might have used that dataset and I think we need to merge the pull request they have their first before we merge this particular pull request. What do you think?

ffl096:
This is a draft pull request, it is not to be merged right now regardless :)

However, just to clarify: I do *not* propose to remove the `coseg` dataset. We have to think about a reasonable data format to deliver the dataset that does not rely on pickle. Ideally, the return value of the `coseg` function should stay exactly the same.
`SimplicialComplex` objects in this pr are compatible to the previous implementation as long as the user does not access internal state. The ICML submissions should all be fine.

mhajij:
> This is a draft pull request, it is not to be merged right now regardless :)
> 
> However, just to clarify: I do _not_ propose to remove the `coseg` dataset. We have to think about a reasonable data format to deliver the dataset that does not rely on pickle. Ideally, the return value of the `coseg` function should stay exactly the same. `SimplicialComplex` objects in this pr are compatible to the previous implementation as long as the user does not access internal state. The ICML submissions should all be fine.

we need to create a Data object to be utilized in the higher order context. I think the one available in torch is good enough.

This is an example on how it can be used in a higher order DL model  https://github.com/pyt-team/TopoModelX/blob/569bd193f81d47e04891376676c034e90cc07554/tutorials/combinatorial/hmc_train.ipynb

mhajij (approved):

ffl096:
The dataset issue still stands and is outside of the scope to be fixed here. We cannot reliably use pickled objects as data objects.

mhajij:
> The dataset issue still stands and is outside of the scope to be fixed here. We cannot reliably use pickled objects as data objects.

I cannot merge wihout passing the tests, what do you think we should do? should we fix the dataset issues first?

mhajij (approved):

ffl096:
According to git blase, the `coseg` dataset downloaded from [here](https://github.com/Ideas-Laboratory/shape-coseg-dataset) was preprocessed by you, right? This repo does not contain this preprocessing script, can you provide that to me? Same for `shrec_16`.

USFCA-MSDS:
@ffl096 What do you want to do with this PR  ? I think we need to have SC faster and implemented correctly but many code relies on the datasets-- what do you suggest?

ffl096:
As outlined above, the dataset structure has to be overhauled completely. This is outside of the scope of this pull request though, and needs to be done regardless. The current system is highly unstable. Once that is done, this pull request is good to be merged.

PR #225, opened 2023-08-23
fix colored hypergraph

mhajij:
some fixes for chg and cc are added. 

(1) incidnce matrix for chg is supposed to be not symmetric eg :  chg.incidence_matrix(1,2) is not equal to chg.incidence_matrix(2,1).
(2) incidence type is only a feature of cc. 
 (3) incidence matrices for both complexes are supposed to receive two indices.


toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+            raise ValueError("incidence must be computed for k!=r, got equal r and k.")
+        if to_rank is None:
+            if incidence_type == "up":
+                children = self.skeleton(rank)
+                uidset = self.skeleton(rank + 1)
+            elif incidence_type == "down":
+                uidset = self.skeleton(rank)
+                children = self.skeleton(rank - 1)
+            else:
+                raise TopoNetXError("incidence_type must be 'up' or 'down' ")
```
ffl096, line 525: This should be a `ValueError`.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+        else:
+            assert (
+                rank != to_rank
+            )  # incidence is defined between two skeletons of different ranks
```
ffl096, line 529: This can be removed. You already checked that case at the beginning of the function.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+        An incidence matrix indexed by r-ranked hyperedges k-ranked hyperedges
+        r !=k, when k is None incidence_type will be considered instead
+
+        Parameters
+        ----------
+        incidence_type : {'up', 'down'}, default='up'
+        sparse : bool, default=True
+        index : bool, default=False
+            If True return will include a dictionary of children uid : row number
+            and element uid : column number
```
ffl096, line 486: There are parameters missing

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+
+        Parameters
+        ----------
+        weight : bool, default=False
+            If False all nonzero entries are 1.
+            If True and self.static all nonzero entries are filled by
+            self.cells.cell_weight dictionary values.
+        index : boolean, optional, default False
+            If True return will include a dictionary of node uid : row number
+            and cell uid : column number
```
ffl096, line 562: Parameters missing.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+        return self._incidence_matrix(
+            rank, to_rank, incidence_type=incidence_type, sparse=sparse, index=index
+        )
+
+    def adjacency_matrix(self, rank, via_rank, s=1, index=False):
+        """Sparse weighted :term:`s-adjacency matrix`.
+
+        Parameters
+        ----------
+        r,k : int, int
```
ffl096, line 581: These paramaters have different names in the function signature.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+            Two ranks for skeletons in the input Colored Hypergraph
+        s : int, list, optional, default : 1
+            Minimum number of edges shared by neighbors with node.
+        index: boolean, optional, default: False
+            If True, will return a rowdict of row to node uid
+        index : book, default=False
```
ffl096, line 587: `index` is documented twice.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+        >>> G = Graph() # networkx graph
+        >>> G.add_edge(0, 1)
+        >>> G.add_edge(0,3)
+        >>> G.add_edge(0,4)
+        >>> G.add_edge(1, 4)
+        >>> CHG = ColoredHyperGraph(cells=G)
+        >>> CHG.adjacency_matrix(0, 1)
+        """
+        if via_rank is not None:
+            assert rank < via_rank
```
ffl096, line 610: Don't use `assert` for this. Raise a proper `ValueError` with meaningful error message please.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+        return A
+
+    def coadjacency_matrix(self, rank, via_rank, s=1, index=False):
+        """Compute the coadjacency matrix.
+
+        The sparse weighted :term:`s-coadjacency matrix`
+
+        Parameters
+        ----------
+        r,k : two ranks for skeletons in the input Colored Hypergraph, such that r>k
```
ffl096, line 662: Other names in the function signature.

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +448,256 @@ def add_cell(self, cell, rank=None, **attr):
...
+            coadjacency_matrix : scipy.sparse.csr.csr_matrix
+
+            row dictionary : dict
+
+        If index if False
+
+            coadjacency_matrix : scipy.sparse.csr.csr_matrix
+        """
+        if via_rank is not None:
+            assert rank > via_rank
```
ffl096, line 688: See above, don't use `assert` for this, please.

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+
+    def is_connected(self, s=1, cells=False):
+        """Determine if combintorial complex is :term:`s-connected <s-connected, s-node-connected>`.
+
+        Parameters
+        ----------
+        s : int, list, optional, default : 1
+            Minimum number of edges shared by neighbors with node.
+
+        cells: boolean, optional, default: False
```
ffl096, line 788: 1. Always use either `optional` or `default: XX` but never both (same for other places in this PR).
2. Its `bool`, not `boolean` in Python.

```suggestion
        cells: bool, default: False
```

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+            If True, will determine if s-cell-connected.
+            For s=1 s-cell-connected is the same as s-connected.
+
+        Returns
+        -------
+        is_connected : boolean
```
ffl096, line 794:
```suggestion
        bool
```

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+
+    def singletons(self):
+        """Return a list of singleton cell.
+
+        A singleton cell is a cell of
+        size 1 with a node of degree 1.
+
+        Returns
+        -------
+        singles : list
```
ffl096, line 824:
```suggestion
        list
```

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+            A list of cells uids.
+        """
+        singletons = []
+        for cell in self.cells:
+            zero_elements = self.cells[cell].skeleton(0)
+            if len(zero_elements) == 1:
+                for n in zero_elements:
+                    if self.degree(n) == 1:
+                        singletons.append(cell)
+        return singletons
```
ffl096, line 834: I think a generator would be a better choice here. If the caller wants a list, they can always just construct one.

```suggestion
        for cell in self.cells:
            zero_elements = self.cells[cell].skeleton(0)
            if len(zero_elements) == 1:
                for n in zero_elements:
                    if self.degree(n) == 1:
                        yield cell
```

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+
+    def remove_singletons(self, name=None):
+        """Construct new CHG with singleton cells removed.
+
+        Parameters
+        ----------
+        name: str, optional, default: None
```
ffl096, line 841:
```suggestion
        name: str, optional
```

toponetx/classes/combinatorial_complex.py
```diff
@@ -528,3 +776,412 @@ def clone(self) -> "CombinatorialComplex":
...
+        """
+        A, coldict = self.cell_adjacency_matrix(s=s, index=True)
+        G = nx.from_scipy_sparse_matrix(A)
+        diams = []
+        comps = []
+        for c in nx.connected_components(G):
+            diamc = nx.diameter(G.subgraph(c))
+            temp = set()
+            for e in c:
+                temp.add(coldict[e])
```
ffl096, line 1052:
```suggestion
            temp = {coldict[e] for e in c}
```

ffl096 (commented):
Only checked for code and documentation style, not for functional issues.

My comments may also apply to other parts of the PR, I commented on a style violation only once even if the same violation also appears on other places. In general:
- A parameter is either `optional` or has a `default=XX` in the parameter documentation. The later already implies that it is optional. Never write both.
- Always use Python's names for primitive data types. It's `bool`, not `boolean`.
- Don't use `assert` for parameter validation. Use an `if`-construct and raise a `ValueError` with appropriate error message. This makes it way easier for a caller to spot what they did wrong.

mhajij:
> ay also apply to other parts of the PR, I commented on a style violation only once even if the same violation also appears on other places. In general:
> 
> * A parameter is either `optional` or has a `default=XX` in the parameter documentation. The later already implies that it is optional. Never write both.

Thank you Florian for the remarks. I am adding Maneel, Guru and Devendra for this conversation. They have been working on CHG and CC and merging them togehter.

@maneel1995 
@gurug-dev 
@devendragovil

ffl096:
Just to add to that point: According to our contribution guidelines, `optional, default: xx` **on a new line** would be the way to go (See [here](https://pyt-team.github.io/topomodelx/contributing/index.html#the-anatomy-of-a-docstring), Point 3a). This, however, is not in line with NumPy's style guide that we are using, see [here](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).

I propose to stick to the later instead of doing our own thing (which we do not follow anywhere in the codebase, strictly speaking), which means that you should use *either* `optional` or `default: xx` and any of both is part of the type definition We should update the contributing guidelines accordingly.

mhajij:
> Just to add to that point: According to our contribution guidelines, `optional, default: xx` **on a new line** would be the way to go (See [here](https://pyt-team.github.io/topomodelx/contributing/index.html#the-anatomy-of-a-docstring), Point 3a). This, however, is not in line with NumPy's style guide that we are using, see [here](https://numpydoc.readthedocs.io/en/latest/format.html#parameters).
> 
> I propose to stick to the later instead of doing our own thing (which we do not follow anywhere in the codebase, strictly speaking), which means that you should use _either_ `optional` or `default: xx` and any of both is part of the type definition We should update the contributing guidelines accordingly.

@maneelusf 
@gurug-dev 
@devendragovil 

Hello everyone, just please keep the above in mind when making your edits

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +447,257 @@ def add_cell(self, cell, rank=None, **attr):
...
+                    )
+
+        self._add_hyperedge(cell, rank=rank, **attr)
+
+    def _incidence_matrix(
+        self,
+        rank,
+        to_rank,
+        incidence_type: Literal["up", "down"] = "up",
+        weight=None,
```
gurug-dev, line 470: @mhajij weight is an unused parameter in the method, this should be removed?

toponetx/classes/combinatorial_complex.py
```diff
@@ -447,9 +447,257 @@ def add_cell(self, cell, rank=None, **attr):
...
+                children = self.skeleton(to_rank)
+                uidset = self.skeleton(rank)
+        return compute_set_incidence(children, uidset, sparse, index)
+
+    def incidence_matrix(
+        self,
+        rank,
+        to_rank=None,
+        incidence_type: str = "up",
+        weight=None,
```
gurug-dev, line 548: same as above, weight parameter is unused.

mhajij:
@gurug-dev test failed after pushing them. Can you please double check ?

gurug-dev (approved):

toponetx/classes/colored_hypergraph.py
```diff
@@ -1,7 +1,7 @@
 """Creation and manipulation of a Colored Hypergraph."""
 
 from collections.abc import Collection, Hashable, Iterable
-from typing import Literal
+from typing import Literal, Optional
```
ffl096, line 4: We are on Python 3.10+, `typing.Optional` is deprecated. Please use union type expressions, e.g. `str | None` instead of `Optional[str]`.

USFCA-MSDS:
this branch is handled elsewhere coloredhg

toponetx/classes/colored_hypergraph.py
gurug-dev: will fix in a separate PR.

PR #236, opened 2023-09-05
Clean up `Complex` ABC

ffl096:
The methods inside `Complex` are abstract and don't have an implementation. The test case was basically only there to make test coverage happy, because the `pass` statement is considered code that is not covered otherwise. We can remove the superfluous `pass` statements instead.


ninamiolane (approved):
Excellent, thanks.

PR #238, opened 2023-09-12
Path Complex

quang-truong:
This pull request contains source code to implement the [Path Complex based on simple paths](https://arxiv.org/pdf/2308.06838.pdf), which is inspired from [the original Path Complex](https://arxiv.org/pdf/1207.2834.pdf).


toponetx/classes/path.py
```diff
@@ -0,0 +1,120 @@
+"""Path class."""
+
+from collections.abc import Collection, Hashable
+from typing import Any, List, Tuple, Union
```
ffl096, line 4: Please don't use the deprecated `typing.List`, `typing.Tuple`, and `typing.Union` classes. TopoNetX is on Python 3.10+, so you can use the builtin `list` and `tuple` classes support subscripting and union type expression (`A | B` instead of `Union[A, B]`) is available.

toponetx/classes/path.py
```diff
@@ -0,0 +1,120 @@
...
+            self._boundaries = self.construct_path_boundaries(
+                elements,
+                reserve_sequence_order=reserve_sequence_order,
+                allowed_paths=allowed_paths,
+            )
+        else:
+            self._boundaries = list()
+
+    @staticmethod
+    def construct_path_boundaries(  # when allowed_paths is not specified, all possible boundaries are constructed similarly to cliques.
```
ffl096, line 61: This shouldn't be a comment but moved to the docstring.

toponetx/classes/path.py
```diff
@@ -0,0 +1,120 @@
...
+        Else, the sub-sequence of nodes in the p-path will be reversed if the first index is larger than the last index.
+    allowed_paths : List[Tuple], optional
+        A list of allowed boundaries. If None, all possible boundaries are constructed (similarly to simplex).
+    attr: keyword arguments, optional
+        Additional attributes to be associated with the p-path.
+    """
+
+    def __init__(
+        self,
+        elements: Union[List, Tuple],
```
ffl096, line 37: The type definition is insufficient:

1. What are the elements of the list or tuple? I suspect `Hashable`? (This applies throughout the pull request)
2. Do you really need a list or a tuple? Probably a protocol from `collections.abc` would be more fitting.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,502 @@
+"""Path complex."""
+from collections.abc import Hashable, Iterable, Iterator
+from itertools import chain
+from typing import List, Set, Tuple, Union
```
ffl096, line 4: See previous comment, you don't need any of these classes from the `typing` module.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,502 @@
...
+        reserve_sequence_order : bool, default=False
+            If True, reserve the order of the sub-sequence of nodes in the p-path. Else, the sub-sequence of nodes in the p-path will
+            be reversed if the first index is larger than the last index.
+        allowed_paths : List[Tuple], optional
+            A list of allowed boundaries. If None, only obvioys boundaries are constructed (sub-sequences where the first or the last index is omitted).
+        max_rank : int, default=3
+            The maximal length of a path in the path complex.
+        attr: keyword arguments, optional
+            Additional attributes to be associated with the path complex.
+        """
```
ffl096, line 50: Document the parameters in the class docstring, see other examples in this package.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,502 @@
...
+        elif isinstance(paths, list) or isinstance(paths, tuple):
+            if len(paths) > 0:
+                paths = [tuple(path) for path in paths]
+                self.add_paths_from(set(paths))
+        elif paths is not None:
+            raise TypeError(
+                "Input paths must be a graph or an iterable of paths as lists or tuples."
+            )
+
+    def add_paths_from(self, paths: Set[Union[List, Tuple, "Path"]]) -> None:
```
ffl096, line 91: `Path` doesn't need quotations here (same throughout this file).

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,502 @@
...
+    def clone(self) -> "PathComplex":
+        """Return a copy of the simplicial complex.
+
+        The clone method by default returns an independent shallow copy of the simplicial complex. Use Python’s
+        `copy.deepcopy` for new containers.
+
+        Returns
+        -------
+        SimplicialComplex
+        """
```
ffl096, line 209: You forgot to update this docstring

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,502 @@
...
+        return PathComplex(self.paths, name=self.name)
+
+    def skeleton(self, rank):
```
ffl096, line 212: Typing annotations are missing.

toponetx/classes/reportviews.py
```diff
@@ -828,3 +826,85 @@ def __contains__(self, e) -> bool:
...
+        return f"PathView({all_paths})"
+
+
+if __name__ == "__main__":
+    view = PathView()
+    view.faces_dict.append({(0,): {"weight": 1}})
+    view.max_dim = 0
+    view.faces_dict.append({(0, 1): {"weight": 1}})
+    view.max_dim = 1
+    print(view)
```
ffl096, line 910: Please remove this.

ffl096 (changes requested):

toponetx/classes/path.py
quang-truong: 1. Yes it should be `Hashable`.
2. I think `Sequence` is appropriate in this case because sequence order matters in path.

toponetx/classes/path.py
quang-truong: Fixed!

toponetx/classes/path.py
```diff
@@ -0,0 +1,162 @@
...
+        A name for the p-path.
+    construct_boundaries : bool, default=False
+        If True, construct the entire boundary of the p-path.
+    reserve_sequence_order : bool, default=False
+        If True, reserve the order of the sub-sequence of nodes in the p-path.
+        Else, the sub-sequence of nodes in the p-path will be reversed if the first index is larger than the last index.
+    allowed_paths : Iterable[tuple[Hashable]], optional
+        A list of allowed boundaries. If None, all possible boundaries are constructed (similarly to simplex).
+    attr: keyword arguments, optional
+        Additional attributes to be associated with the p-path.
```
mhajij, line 32: Define a path mathemtically, put a few easy to follow examples here.

toponetx/classes/path.py
```diff
@@ -0,0 +1,162 @@
...
+        else:
+            self._boundaries = list()
+
+    @staticmethod
+    def construct_path_boundaries(
+        elements: Sequence[Hashable],
+        reserve_sequence_order: bool = False,
+        allowed_paths: Iterable[tuple[Hashable]] = None,
+    ) -> list[tuple[Hashable]]:
+        """
```
mhajij, line 66: define this method mathematically, give an example. Make it easy and pythonic.

toponetx/classes/path.py
```diff
@@ -0,0 +1,162 @@
...
+                and len(boundary) > 1
+                and boundary[0] > boundary[-1]
+            ):
+                boundary.reverse()
+            if allowed_paths is None or tuple(boundary) in allowed_paths:
+                boundaries.append(tuple(boundary))
+        return boundaries
+
+    @property
+    def boundary(self) -> list[tuple[Hashable]]:
```
mhajij, line 98: define what you mean by boundary here. Note, the term boudary matrix is not used in toponetx, instead we are using incidence matrix. If this is what you mean then you probably need to change it to be consistent

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,657 @@
...
+        A name for the path complex.
+    reserve_sequence_order : bool, default=False
+        If True, reserve the order of the sub-sequence of nodes in the p-path. Else, the sub-sequence of nodes in the p-path will
+        be reversed if the first index is larger than the last index.
+    allowed_paths : Iterable[tuple[Hashable]], optional
+        An iterable of allowed boundaries. If None, only obvious boundaries are constructed (sub-sequences where the first or the last index is omitted).
+    max_rank : int, default=3
+        The maximal length of a path in the path complex.
+    attr: keyword arguments, optional
+        Additional attributes to be associated with the path complex.
```
mhajij, line 36: give a mathematical definition of path complex here. Make it simple and give 2 examples that are easy to understand.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,657 @@
...
+        paths : Iterable[Sequence[Hashable] or Path]
+            an iterable of paths as lists, tuples, or Path objects.
+        """
+        if isinstance(paths, Hashable):
+            raise TypeError("Paths must be a iterable of paths as lists, tuples.")
+        paths_clone = paths.copy()
+        for p in paths_clone:
+            self.add_path(p)
+
+    def add_path(self, path: Hashable | Sequence[Hashable] | Path, **attr) -> None:
```
mhajij, line 104: define a  path mathmetically. Remember that the user does not have to know path complex very well to be able to use this method so an easy example should guide the user to create a path without much effort and hence utilize the complex also without much effort.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,657 @@
...
+                    path_dict,
+                    abs(boundary),
+                )
+        else:
+            if signed:
+                return boundary
+            else:
+                return abs(boundary)
+
+    def coincidence_matrix(self, rank: int, signed: bool = True, index: bool = False):
```
mhajij, line 378: Since this matrix is just the transponse of the earlier one then I do not think it needed as a separate method. WE did not include these methods in other complexes.

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,657 @@
...
+    def __str__(self) -> str:
+        """Return detailed string representation."""
+        return f"Path Complex with shape {self.shape} and dimension {self.dim}"
+
+    def __repr__(self) -> str:
+        """Return string representation."""
+        return f"PathComplex(name='{self.name}')"
+
+    @staticmethod
+    def compute_allowed_paths(
```
mhajij, line 602: what do you mean by allowed ? give example and give a definition.

mhajij (changes requested):

toponetx/classes/path_complex.py
quang-truong: I just removed the `coincidence_matrix()`. I notice that `SimplicialComplex()` still has this method.

toponetx/classes/path_complex.py
quang-truong: Just added.

toponetx/classes/path_complex.py
quang-truong: Just added.

toponetx/classes/path.py
quang-truong: Boundary here conveys the same meaning with the boundary method in `simplex.py` and `cell.py`. I just want `path.py` to follow the similar interface.

toponetx/classes/path.py
quang-truong: Just added.

mhajij:
Please add the following methods :

(1) to_other_complex()

where other_complex is any more general of path complex. e.g colored HG and combinatorial complex.

(2) restrict_to_nodes

(3) restrict to paths 

(4) set attrs and get attrs (similar to https://github.com/pyt-team/TopoNetX/blob/da155dfbb46a77a9034a1d5707b573f16b93cce9/toponetx/classes/cell_complex.py#L957)

(5) add hodge_laplacian_spectrum methods in the file  :  


https://github.com/pyt-team/TopoNetX/blob/main/toponetx/algorithms/spectrum.py


(6) add tutorial on path complex demonstrating the main methods.

toponetx/classes/path.py
mhajij: sure, we should define that in these classes as well. Please raise an issue on that.

quang-truong: I mean `boundary()` already exists in `simplex.py` and `cell.py`, so no need to raise an issue on this.

https://github.com/pyt-team/TopoNetX/blob/da155dfbb46a77a9034a1d5707b573f16b93cce9/toponetx/classes/simplex.py#L100-L107

https://github.com/pyt-team/TopoNetX/blob/da155dfbb46a77a9034a1d5707b573f16b93cce9/toponetx/classes/cell.py#L205-L215

mhajij:
Finally for adj/incidence/laplacian matrices please add weight parameter to make the signature consistent across the package #242.

quang-truong:
> Please add the following methods :
> 
> (1) to_other_complex()
> 
> where other_complex is any more general of path complex. e.g colored HG and combinatorial complex.
> 
> (2) restrict_to_nodes
> 
> (3) restrict to paths
> 
> (4) set attrs and get attrs (similar to
> 
> https://github.com/pyt-team/TopoNetX/blob/da155dfbb46a77a9034a1d5707b573f16b93cce9/toponetx/classes/cell_complex.py#L957
> 
> )
> (5) add hodge_laplacian_spectrum methods in the file :
> 
> https://github.com/pyt-team/TopoNetX/blob/main/toponetx/algorithms/spectrum.py
> 
> (6) add tutorial on path complex demonstrating the main methods.

Regarding (1), the current implementation of `CombinatorialComplex` prevents me from generalizing `PathComplex`, because paths take into account order sequence, while `CombinatorialComplex` converts all inputs to `frozenset`.

Besides that, all of the above methods are implemented. I am working on (6), which will take a while. \

mhajij:
@ffl096 can you please take a look at this one when you get a chance, give your comments ? Thanks!

mhajij:
@ninamiolane any high level comments?

toponetx/algorithms/spectrum.py
```diff
@@ -255,7 +256,27 @@ def simplicial_complex_hodge_laplacian_spectrum(
     return laplacian_spectrum(SC.hodge_laplacian_matrix(rank=rank))
 
 
-def cell_complex_adjacency_spectrum(CC: CellComplex, rank):
+def path_complex_hodge_laplacian_spectrum(
+    PC: PathComplex, rank: int, weight: str = "weight"
```
ffl096, line 260: According to the docstring, `weight` is optional:
```suggestion
    PC: PathComplex, rank: int, weight: str | None = "weight"
```

toponetx/algorithms/spectrum.py
```diff
@@ -255,7 +256,27 @@ def simplicial_complex_hodge_laplacian_spectrum(
...
+):
+    """Return eigenvalues of the Laplacian of PC.
+
+    Parameters
+    ----------
+    PC : PathComplex
+    rank : int
+    weight : str or None, optional (default='weight')
```
ffl096, line 268:
```suggestion
    weight : str or None, default='weight'
```

toponetx/algorithms/spectrum.py
```diff
@@ -255,7 +256,27 @@ def simplicial_complex_hodge_laplacian_spectrum(
...
+        If None, then each cell has weight 1.
+
+    Returns
+    -------
+    evals : NumPy array
```
ffl096, line 273: Single return values do not have a name according to NumPy's documentation style.

```suggestion
    np.ndarray
```

toponetx/algorithms/spectrum.py
```diff
@@ -305,6 +326,26 @@ def simplicial_complex_adjacency_spectrum(
...
+    Parameters
+    ----------
+    PC : PathComplex
+    dim : int
+    weight : str, optional
+        If None, then each cell has weight 1.
+
+    Returns
+    -------
+    evals : NumPy array
```
ffl096, line 343: Single return values do not have a name according to NumPy's documentation style.

```suggestion
    np.ndarray
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,218 @@
...
+                allowed_paths=allowed_paths,
+            )
+        else:
+            self._boundaries = list()
+
+    @staticmethod
+    def construct_path_boundaries(
+        elements: Sequence[Hashable],
+        reserve_sequence_order: bool = False,
+        allowed_paths: Iterable[tuple[Hashable]] = None,
```
ffl096, line 98:
```suggestion
        allowed_paths: Iterable[tuple[Hashable]] | None = None,
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,218 @@
...
+    [(2, 3), (1, 3)]
+    """
+
+    def __init__(
+        self,
+        elements: Sequence[Hashable],
+        name: str = "",
+        construct_boundaries: bool = False,
+        reserve_sequence_order: bool = False,
+        allowed_paths: Iterable[tuple[Hashable]] = None,
```
ffl096, line 73:
```suggestion
        allowed_paths: Iterable[tuple[Hashable]] | None = None,
```

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,1149 @@
...
+from toponetx.classes.combinatorial_complex import CombinatorialComplex
+from toponetx.classes.complex import Complex
+from toponetx.classes.path import Path
+from toponetx.classes.reportviews import PathView
+
+__all__ = ["PathComplex"]
+
+
+class PathComplex(Complex):
+    """A class representing a path complex based on simple paths as proposed in (https://arxiv.org/abs/2308.06838). The original path complex is defined in (https://arxiv.org/pdf/1207.2834.pdf).
```
ffl096, line 21: Please use proper references (Reference section) for this. See for example the modules in TopoModelX

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,1149 @@
...
+    >>> G = nx.Graph()
+    >>> G.add_edges_from([(1, 2), (2, 3), (1, 3)])
+    >>> PC = PathComplex(G)
+    >>> PC.paths
+    PathView([(1,), (2,), (3,), (1, 2), (1, 3), (2, 3), (1, 3, 2), (1, 2, 3), (2, 1, 3)])
+    """
+
+    def __init__(
+        self,
+        paths: nx.Graph | Iterable[Sequence[Hashable]] = None,
```
ffl096, line 78:
```suggestion
        paths: nx.Graph | Iterable[Sequence[Hashable]] | None = None,
```

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,1149 @@
...
+        name: str = "",
+        reserve_sequence_order: bool = False,
+        allowed_paths: Iterable[tuple[Hashable]] = None,
```
ffl096, line 81:
```suggestion
        allowed_paths: Iterable[tuple[Hashable]] | None = None,
```

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,1149 @@
...
+        --------
+        >>> PC = PathComplex([(1, 2, 3)])
+        >>> PC.paths
+        PathView([(1,), (2,), (3,), (1, 2), (2, 3), (1, 2, 3)])
+        >>> PC.add_paths_from([(1, 2, 4), (1, 2, 5), (4,5)])
+        >>> PC.paths
+        PathView([(1,), (2,), (3,), (4,), (5,), (1, 2), (2, 3), (2, 4), (2, 5), (4, 5), (1, 2, 3), (1, 2, 4), (1, 2, 5)])
+        """
+        if isinstance(paths, Hashable):
+            raise TypeError("Paths must be a iterable of paths as lists, tuples.")
```
ffl096, line 165: Please remove this check

toponetx/classes/path_complex.py
```diff
@@ -0,0 +1,1149 @@
...
+        Returns
+        -------
+        set[tuple[Hashable]]
+            Set of elementary p-paths of dimension specified by `rank`.
+        """
+        if rank < len(self._path_set.faces_dict) and rank >= 0:
+            tmp = (path for path in self._path_set.faces_dict[rank].keys())
+            return sorted(
+                tmp, key=lambda x: tuple(map(str, x))
+            )  # lexicographic comparison
```
ffl096, line 342: This doesn't conform with the function signature and docstring.

ffl096 (changes requested):

toponetx/classes/path_complex.py
mhajij: @quang-truong I am not sure if you have addressed this one, if so please mark it as resolved. Thanks!

quang-truong: @mhajij @ffl096 i just updated the code for `skeleton()` and added extra test cases. Please refer to the latest commit for changes.

toponetx/algorithms/spectrum.py
```diff
@@ -262,7 +263,29 @@ def simplicial_complex_hodge_laplacian_spectrum(
...
+    weight : str or None, default='weight'
+        If None, then each cell has weight 1.
+
+    Returns
+    -------
+    np.ndarray
+        Eigenvalues.
+    """
+    return laplacian_spectrum(
+        PC.hodge_laplacian_matrix(rank=rank, signed=True), weight=weight
```
ffl096, line 284: Shouldn't the `weight` attribute be passed to `PC.hodge_laplacian_matrix` instead? While `laplacian_spectrum` also has a `weight` attribute, it is never used there (and should be removed).

toponetx/algorithms/spectrum.py
```diff
@@ -262,7 +263,29 @@ def simplicial_complex_hodge_laplacian_spectrum(
...
+    )
+
+
+def cell_complex_adjacency_spectrum(CCC: CellComplex, rank):
```
ffl096, line 288: - Missing type for `rank`.
- Cell complexes are abbreviated `CC`, not `CCC`.

```suggestion
def cell_complex_adjacency_spectrum(CC: CellComplex, rank: int):
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+        self,
+        elements: Sequence[Hashable],
+        name: str = "",
+        construct_boundaries: bool = False,
+        reserve_sequence_order: bool = False,
+        allowed_paths: Iterable[tuple[Hashable]] | None = None,
+        **attr,
+    ) -> None:
+        self.__check_inputs(elements, reserve_sequence_order)
+        if isinstance(elements, int) or isinstance(elements, str):
```
ffl096, line 86: This cannot happen according to the type definition of `elements`. Either support this explicitly or remove this (I would prefer the later).

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+            The nodes in the elementary p-path.
+        reserve_sequence_order : bool, default=False
+            If True, reserve the order of the sub-sequence of nodes in the elementary p-path.
+            Else, the sub-sequence of nodes in the elementary p-path will be reversed if the first index is larger than the last index.
+        allowed_paths : Iterable[tuple[Hashable]], optional
+            A list of allowed boundaries. If None, all possible boundaries are constructed (similarly to simplex).
+
+        Returns
+        -------
+        boundaries : list[tuple[Hashable]]
```
ffl096, line 124: Single return values do not have a name.

```suggestion
        list[tuple[Hashable]]
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+        return boundaries
+
+    @property
+    def boundary(self) -> list[tuple[Hashable]]:
+        """
+        Return list of elementary (p-1)-path objects representing the elementary (p-1)-path on the boundary of the target elementary p-path.
+
+        Returns
+        -------
+        boundaries : list[tuple[Hashable]]
```
ffl096, line 161:
```suggestion
        list[tuple[Hashable]]
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+            A shallow copy of the elementary p-path.
+        """
+        return Path(
+            self.elements,
+            name=self.name,
+            construct_boundaries=self.construct_boundaries,
+            **self._properties,
+        )
+
+    def __check_inputs(self, elements: Any, reserve_sequence_order: bool):
```
ffl096, line 182:
```suggestion
    def __check_inputs(self, elements: Any, reserve_sequence_order: bool) -> None:
```

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+        ----------
+        elements: Any
+            The nodes in the elementary p-path.
+        reserve_sequence_order : bool
+            If True, reserve the order of the sub-sequence of nodes in the elementary p-path.
+            Else, the sub-sequence of nodes in the elementary p-path will be reversed if the first index is larger than the last index.
+            This variable is only used for sanity check on the input.
+        """
+        if isinstance(elements, Sequence):
+            if not isinstance(elements, list) and not isinstance(elements, tuple):
```
ffl096, line 196: Does it have to be a `list` or `tuple`? I would prefer to use the broadest possible protocol in `collections.abc`.

toponetx/classes/path.py
```diff
@@ -0,0 +1,227 @@
...
+                        elements
+                    )
+                )
+        elif isinstance(elements, Hashable):
+            if isinstance(elements, int) or isinstance(elements, str):
+                pass
+            else:
+                raise ValueError(
+                    f"Elements of an elementary p-path must be a list or tuple, got {type(elements)}"
+                )
```
ffl096, line 219:
```suggestion
            if not isinstance(elements, int) and not isinstance(elements, str):
                raise ValueError(
                    f"Elements of an elementary p-path must be a list or tuple, got {type(elements)}"
                )
```

Plus: The error message doesn't seem to correctly describe the problem here.

toponetx/classes/reportviews.py
```diff
@@ -828,3 +826,95 @@ def __contains__(self, e) -> bool:
...
+    Parameters
+    ----------
+    name: str, optional
+        Name of the PathView instance.
+    """
+
+    def __init__(self, name: str = "") -> None:
+        super().__init__(
+            name
+        )  # self.faces_dict is a list of dictionary of allowed paths
```
ffl096, line 844: The comment doesn't seem to belong here?

ffl096 (changes requested):
Some things I didn't notice before, sorry! Other than that, this pull request looks good to me :)

toponetx/classes/path.py
mhajij: I think paths have to be ordered, so sets or frozensets do not work.

ffl096: There are many other data structures, that's why the protocols in `collections.abc` exist. I don't see why e.g., `collections.abc.Sequence` wouldn't be enough? All sequences are ordered.

toponetx/classes/reportviews.py
quang-truong: It was a comment when I first wrote the code so I didn't forget what should be in `self.faces_dict`. Anyway I just removed it :)

toponetx/classes/path.py
quang-truong: Yes just replaced it with a more accurate error message.

toponetx/classes/path.py
quang-truong: Yes you're correct on this. `collections.abc.Sequence` is sufficient.

toponetx/classes/path.py
quang-truong: I already have some test cases covering this, so I still keep and explicitly support it in function signature. `__check_inputs()` also supports this from the beginning. My intuition is that users may prefer to explitcitly call `Path(3)` instead of `Path([3])` when instantiating a 0-path.

toponetx/algorithms/spectrum.py
quang-truong: just removed `weight` from `laplacian_spectrum()` and moved `weight` to `PC.hodge_laplacian_matrix`

ffl096 (approved):

mhajij:
@quang-truong we are aiming to publish the package soon and the publisher requires almost 100% code coverage, can you try to increase the code coverage of path complex to its maximal ? thank you

mhajij (approved):

ninamiolane (approved):
Excellent, thank you very much for this contribution!

PR #244, opened 2023-09-19
Deprecate simplex tree construction in `Simplex`.

ffl096:
This pull request deprecates `Simplex.boundary` and `Simplex.faces`. As a simplex has no knowledge of the SC it belongs to, the sub-simplices returned by both functions lack any user-defined parameters that were given when constructing the simplicial complex. Going through appropriate methods in `SimplicialComplex` avoids this problem.

With the deprecation of the above functions, I also removed the simplex tree construction inside of the `Simplex` class. The default value `construct_tree=True` meant a significant performance implication.
In fact, we have fallen into this trap ourself inside TopoNetX at least once: The `Simplex` construction in `SimplicialComplex.__init__` does not set this parameter to false even though it definitely should have been.

### DIFF
--- a/test/classes/test_simplex.py
+++ b/test/classes/test_simplex.py
@@ -13,9 +13,11 @@ class TestSimplex:
         s = Simplex((1,))
         assert len(s) == 1
         assert tuple(s) == (1,)
-        assert s.construct_tree is True
         assert s.name == ""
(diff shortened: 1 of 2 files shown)


ninamiolane (approved):
Nice, thanks!

PR #247, opened 2023-09-19
Clean-up abbreviations of cell and combinatorial complexes

ffl096:
See #95.


ninamiolane (approved):
Wow, so fast! Thank you so much.

PR #261, opened 2023-09-28
fixing adjacency matrix issues

mhajij:


ffl096:
This added a build directory that shouldn’t be in the repository.

mhajij:
> This added a build directory that shouldn’t be in the repository.

which directory ?

PR #264, opened 2023-10-04
Clean-up exception classes

### LINKED ISSUE #260: Clean-up exception classes used in the package
Almost everywhere we use the [built-in exceptions](https://docs.python.org/3/library/exceptions.html) from the standard library (e.g., `KeyError` and `ValueError`). We however also ship three [own exception classes](https://github.com/pyt-team/TopoNetX/blob/44a60d09fc2e8b5cf8c91b669b61021c296b34ca/toponetx/exception.py) that are sporadically used.

In its current form, this is not optimal. First of all this should be consistent, second and more importantly, our own exceptions only inherent from `Exception` and not from the correct built-in subclasses. `TopoNetXNotImplementedError` for example should inherit from the built-in `NotImplementedError`.

# Tasks

- [x] Remove our own classes where one of the built-in exceptions is more appropriate.
- [x] Check that we use the correct errors throughout the package. For example if a parameter type is incorrect, raise a `TypeError` and not a `ValueError`.

ffl096:
closes #260


ninamiolane (approved):
Thanks!

Issue #266, opened 2023-10-05
Inconsistencies in iterating over different complex types


ffl096:
Iterating over complexes is inconsistent across different complex types:
- Iterating over a simplicial or path complex yields all simplices/paths.
- Iterating over a cell complex or coloured hypergraph yields only the nodes.
- Iterating over a combinatorial complex is not implemented at all.

We should make this consistent across all implementations and add a definition to the `Complex` ABC. I would suggest for the `__iter__` method to iterate over all atoms in a complex. Any thoughts?

ninamiolane:
Thanks for spotting this! I agree to iterate only over the nodes, great idea.

sanity check: Is it what networkx does?

ffl096:
NetworkX iterates over the nodes.

I think there is a need for both: Iterating over the nodes and over all faces in the complex. The normal `__iter__` method could iterate over the nodes and a dedicated `iter_atoms` over all faces in the complex.
This would be akin to NetworkX's iterating over `G` (or `G.nodes`) and over `G.edges`.

PR #267, opened 2023-10-06
fixing iter for complexes and making them all behave the same way

mhajij:
This is addressing #266 in one way. 

__iter__ now iterates over the nodes of the complex, consistent with NetworkX. 

complex.cells, complex.paths, complex.simplices can be used to iterate over the atoms in each class.

combinatorial complex inherets from chg and no need to have its own iter method.

### DIFF
--- a/toponetx/classes/path_complex.py
+++ b/toponetx/classes/path_complex.py
@@ -1073,7 +1073,7 @@ class PathComplex(Complex):
         -------
         dict_keyiterator
         """
-        return chain.from_iterable(self._path_set.faces_dict)
+        return iter(self.nodes)
 
     def __len__(self) -> int:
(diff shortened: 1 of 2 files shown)


ffl096 (approved):

ffl096:
Can you add an abstract method to the `Complex` ABC that ensures that `__iter__` is implemented and that defines (in the docstring) that iterating will happen over the nodes? Otherwise this seems good to go!

ffl096:
> Can you add an abstract method to the `Complex` ABC that ensures that `__iter__` is implemented and that defines (in the docstring) that iterating will happen over the nodes? Otherwise this seems good to go!

I just noticed that this is already the case, so simplicial and path complexes violated that contract all along. This is good to go then!

Issue #256, opened 2023-09-22
Increasing code coverage for toponetx to 99%


mhajij:
### What? 

Increase the code coverage for TopoNetX, making it ready for publications.

### How ?

The following classes require attentions as they have many methods that remain untested:

* ColoredHyperGraph
* CombintorialComplex
* CellComplex
* reportview
* Simplex

ninamiolane:
This complements issue #246

mhajij:
@gurug-dev any update on this ?

gurug-dev:
@mhajij will be adding a PR this weekend to address this.

ninamiolane:
We are not at 99% but we are very close, and I think it is enough for now. Closing it, awesome work everyone!

PR #269, opened 2023-10-06
Increasing test coverage for TopoNetX (3)

gurug-dev:
Adding test cases for few exceptions in newly added algorithms (distance, cell_distance, distance_measures) as part of #256


ninamiolane (approved):
Nice!! 🚀

PR #274, opened 2023-10-10
Improve comprehensions and enforce codestyle

ffl096:
Adds the [`flake8-comprehensions`](https://github.com/adamchainz/flake8-comprehensions) package and cleans up reported problems. See the packages documentation for the list of rules and their reasons.


ninamiolane (approved):
What a great idea to use flake8-comprehensions! Thank you for this.

PR #275, opened 2023-10-10
Add atomlist to store and read complexes

ffl096:
Introduce functions for "atom lists", which work similar to edge lists and allow to store and load complexes.

Background: Right now, we ship some datasets in form of pickled data structures. While this was a fast way to implement this problem, it is a really bad idea for various reasons. It is highly unstable as it relies on internal state of data structures and any change to them breaks the datasets. In fact, this is what blocks #220, which completely overhauls the internals of `SimplicialComplex`.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -22,3 +22,4 @@ The API reference gives an overview of `TopoNetX`, which consists of several mod
    algorithms
    generators
    transform
+   readwrite
--- a/docs/api/readwrite.rst
+++ b/docs/api/readwrite.rst
@@ -1,0 +1,9 @@
+*********
+Read-Write
+*********
+
+.. automodule:: toponetx.readwrite.atomlist
+    :members:
+
+.. automodule:: toponetx.readwrite.serialization
+    :members:
--- a/test/readwrite/test_atomlist.py
+++ b/test/readwrite/test_atomlist.py
@@ -1,0 +1,140 @@
+"""Tests for the atomlist read/write functions."""
+import pytest
+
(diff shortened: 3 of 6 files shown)


ffl096:
This is ready for review now.

In a following step, we should migrate datasets relying on pickle to other formats.

ffl096:
I am not exactly sure how to deal with the documentation issues raised my numpydoc. It complains about a lack of documentation for `@overload` functions. These are only there for mypy (and other static type checkers / IDE support) and do not appear in our documentation and are never called.

Thus I would think that they should not have a docstring, the function is documented in the "actual" implementation (the last without `@overload`). Can we skip overloaded functions for numpydoc somehow?

CC @devendragovil

devendragovil:
HI @ffl096 

Yes we can do that, see karate_club at: https://github.com/pyt-team/TopoNetX/blob/main/toponetx/datasets/graph.py

Essentially just leave an inline comment `# numpydoc ignore=GL08`

I also raised an issue #326 where majority of the docstrings issues are due to overloaded methods, maybe we can tick them off as well if we are planning to skip them.

ffl096:
Yes we should ignore them. I was just wondering if we can automatically do that for all overloaded functions ([as done for code coverage](https://github.com/pyt-team/TopoNetX/blob/2dbc4db3cc54fc43e856d5f5ff0e2d7c78dafd91/pyproject.toml#L124)), but didn't found a way to do that. Adding the ignore every time is also fine though.

michaelschaub (approved):

PR #278, opened 2023-10-12
adding dirac operator

IJabea:
Adding dirac operator 

https://arxiv.org/abs/2301.10137


ninamiolane (changes requested):
Missing tests. Every PR should come with its unit-tests.

mhajij (approved):

mhajij (approved):

ffl096 (changes requested):
I don't like the test cases to be honest. While it satisfies code coverage, it really doesn't test that the result is correct. Can we please add an explicit (hand-calculated) result for a simple example and assert that the function returns exactly this matrix?

ninamiolane (changes requested):
I agree with @ffl096 💯

PR #304, opened 2023-11-02
adding random cell complex generator

mhajij:
I do not like the name of the function used in in cc random generator

### DIFF
--- a/test/generators/test_random_cell_complexes - Copy.py
+++ b/test/generators/test_random_cell_complexes - Copy.py
@@ -8,16 +8,15 @@ class TestCcNpComplex:
     def test_zero_nodes(self):
         """Test `cc_np_complex` for zero nodes."""
         cc = cc_np_complex(0, 0.1)
-        assert cc.shape == (0,0,0)
+        assert cc.shape == (0, 0, 0)
 
     def test_non_zero_nodes(self):
         """Test `cc_np_complex` for nonzero nodes."""
         cc = cc_np_complex(5, 0.1)
         assert cc.shape[0] == 5
-        
-    def test_non_zero_nodes(self):
+
+    def test_zero_probability(self):
         """Test `cc_np_complex` for zero probability."""
         cc = cc_np_complex(40, 0)
         assert cc.shape[0] == 40
-
         assert cc.shape[-1] == 0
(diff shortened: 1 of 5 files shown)


ffl096:
We should come up with a better naming scheme. The function name is weird to me. In particular, "complex" is twice in the name, first as part of "cc" and as suffix at the end.

toponetx/generators/random_cell_complexes.py
```diff
@@ -0,0 +1,34 @@
...
+
+from toponetx.classes import CellComplex
+from toponetx.transform.graph_to_cell_complex import homology_cycle_cell_complex
+
+__all__ = [
+    "cc_np_complex",
+]
+
+
+def cc_np_complex(n: int, p: float, seed=None) -> CellComplex:
```
devendragovil, line 13: seed=None is invalid default condition as it will raise an exception, when the user doesn't provide a seed.

test/generators/test_random_cell_complexes - Copy.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def test_zero_nodes(self):
+        """Test `cc_np_complex` for zero nodes."""
+        cc = cc_np_complex(0, 0.1)
+        assert cc.shape == (0, 0, 0)
+
+    def test_non_zero_nodes(self):
+        """Test `cc_np_complex` for nonzero nodes."""
+        cc = cc_np_complex(5, 0.1)
+        assert cc.shape[0] == 5
```
devendragovil, line 16: Ideally a test case for initialization without the second argument (seed) must be added.

devendragovil (changes requested):

test/generators/test_random_cell_complexes.py
```diff
@@ -0,0 +1,22 @@
+"""Test the random cell complex generators."""
+from toponetx.generators.random_cell_complexes import np_cell_complex
+
+
+class TestCcNpComplex:
```
ffl096, line 5: Name is not correct.

ffl096 (changes requested):
Please also update `docs/api/generators.rst` to generate documentation for the new module.

devendragovil (commented):
LGTM, I have checked the docstrings in particular for compliance with numpydoc.

devendragovil (approved):
LGTM. The docs file has also been updated.

Issue #250, opened 2023-09-20
Change thumbnail of ColoredHypergraph tutorials on doc website


ninamiolane:
# What?

Every tutorial on toponetx's doc website has a thumbnail. Currently, the tutorial on colored hypergraphs uses a hypergraph thumbnail. This should be corrected with a colored hypergraph thumbnail.

# Why?

Otherwise, it is incorrect and confusing for users who might mix up hypergraphs and colored hypergraphs.

devendragovil:
@mhajij  I would like your advice on how should I visually depict a colored hypergraph.

devendragovil:
@ninamiolane @mhajij I am thinking of how would you like me to depict the colored hypergraph. Would just shading some hypergraph cells with different colors be sufficient?

mhajij:
@devendragovil I'll send you something later today

devendragovil:
@mhajij @maneelusf Please send me the thumbnail for the CHG tutorial.

ninamiolane:
@mhajij @maneelusf @devendragovil what is the update here?

USFCA-MSDS:
@maneelusf can you please close this issue as soon as possible?

devendragovil:
This has been completed by changing the thumbnail.

PR #307, opened 2023-11-08
Changing thumbnail of CHG

maneelusf:
This is for Issue #250


maneelusf:
@mhajij , please suggest changes as necessary.

USFCA-MSDS (approved):

PR #324, opened 2023-11-16
Simplicial Complex docstrings to reduce Sphnix.

maneelusf:


ffl096 (approved):

toponetx/classes/simplicial_complex.py
```diff
@@ -136,22 +147,35 @@ def shape(self) -> tuple[int, ...]:
         Returns
         -------
         tuple of ints
```
devendragovil, line 149: Since we are following type hint, the proper syntax would be somthing like tuple[int, int] for a tuple of two ints or tuple[int, ...] for variable number of ints in the tuple. See https://docs.python.org/3/library/typing.html#annotating-tuples

toponetx/classes/simplicial_complex.py
```diff
@@ -258,31 +307,61 @@ def __len__(self) -> int:
         return len(self.skeleton(0))
 
     def __getitem__(self, simplex):
-        """Get simplex."""
+        """Get the data associated with the given simplex.
+
+        Parameters
+        ----------
+        simplex : tuple[int, ...]
+            The simplex to retrieve.
```
devendragovil, line 315: See this for reference. This is right.

toponetx/classes/simplicial_complex.py
```diff
@@ -258,31 +307,61 @@ def __len__(self) -> int:
...
 
     def __iter__(self) -> Iterator:
-        """Iterate over all faces of the simplicial complex.
+        """Iterate over all simplices (faces) of the simplicial complex.
 
         Returns
         -------
-        dict_keyiterator
+        Iterator[Tuple[int, ...]]
+            An iterator over all simplices in the simplicial complex.
```
devendragovil, line 338: Tuple should be small case tuple

devendragovil (changes requested):
Requesting some changes.

devendragovil:
Hi, @ffl096, is it possible to revert, to enforce some changes as suggested?

toponetx/classes/simplicial_complex.py
ffl096: This is actually as per NumPy's recommendation. I know we mix the two right now, but neither is wrong. If we want to unify this, that should be a separate PR fixing it everywhere and not just here :)

ffl096:
> Hi, @ffl096, is it possible to revert, to enforce some changes as suggested?

It's not (easily). I don't think we should bother just for a change in capitalisation. A follow-up PR would be great though.

devendragovil:
Sure, I will resolve the conversation.

toponetx/classes/simplicial_complex.py
devendragovil: Yeah, resolving for now. I have a personal preference for type hint one.

PR #332, opened 2024-01-23
Remove hard requirement for `hypernetx`.

ffl096:
The requirement on HyperNetX is not hard and is problematic, as it is not compatible with NetworkX 3, which is more than a year old. With this pull request, users that do not use hypergraphs are free to use NetworkX 3, while others must stay on NetworkX 2 (as transient dependency due to HyperNetX).

We exclusively use HyperNetX in places where we transform user-provided hypergraph instances into complex classes or vice versa. Such code points are only ever called when the user itself interacts with HyperNetX and thus did install it themself. We don't need to force that dependency.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.10", "3.11"]
+        flavor: ["dev", "all"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


ffl096:
For why the hard dependency on HyperNetX and its incompatibility with NetworkX 3 is problematic, see e.g. here: https://github.com/pyt-team/TopoEmbedX/issues/39#issuecomment-1906217924. It now forces us to stay with older versions of SciPy as well.

PR #336, opened 2024-02-06
A simplicial complex has dimension equal to the size of its maximal s…

rballeba:
A simplicial complex has dimension equal to the size of its maximal set minus one. If, for example, a simplicial complex has dimension one, then, it has simplices of at most two vertices, with at least one having exactly two vertices. Currently, the method graph_to_clique_complex was considering that the dimension of a simplex is exactly the number of vertices in it. This commit solves this problem. Also, it only copies the edge attributes from the networkx graph if the resulting simplicial complex has dimension greater or equal than one.


ffl096:
I'm on the fence on this one. The _dimension_ that is meant here is the sizes of the cliques, not of the simplicial complex. We call the later one _rank_ throughout the code base.

rballeba:
As I user, I would expect max_dim to correspond to the dim of the object I am generating. Is there any other use of the term dimension for graphs that coincides with the length and not the length minus one in the literature? @mhajij

USFCA-MSDS:
> As I user, I would expect max_dim to correspond to the dim of the object I am generating. Is there any other use of the term dimension for graphs that coincides with the length and not the length minus one in the literature? @mhajij
> 
> By the way, the pull request is also solving a problem when the user is requesting to generate a point cloud. It is not a usual thing, but it can happen.

@rballeba  @ffl096 
We are using both terms : dimension means the dimension of the of the complex : https://github.com/pyt-team/TopoNetX/blob/7b8fa8b4ff56571a53999823b97ca78e20c7e86c/toponetx/classes/cell_complex.py#L229

whereas rank identifies a the dimenion of a particular skeleton in the complex. In other words :  0=< rank<= max_dim

rballeba:
@USFCA-MSDS Therefore, this means that we can merge this pull request right? @ffl096

ffl096:
My point above still stands: If we want to change the behaviour, the parameter must be renamed. As of right now, this is a hidden break of any consumer code (existing code will run without any error but with changed result!).

The docstring should be updated as well, it still describes the current behaviour.

ffl096:
> This is what was wrong before. Could you point me what should I change from the docstring?

It was not. It specifically talks about the dimensions of _cliques_. I see that this may be confusing and hence I am not opposed to this change. However, if the function should be about the simplex rank, this has to be stated there.

> Regarding changing the code. I'm not a usual maintainer, so I would prefer to avoid larger changes in the codebase.

Not sure what you mean by larger changes to the codebase? Its just about renaming the parameter.

rballeba:
@ffl096 I have not found any standard definition of dimension of a clique. Could you point to any reference where this term is used explicitly? If we see the clique as a simplex, then it has dimension equal to the number of vertices minus one. 

About renaming the parameter, you mean renaming the parameter of the function or all the parameters related to the dimension of cliques?

ffl096:
> About renaming the parameter, you mean renaming the parameter of the function or all the parameters related to the dimension of cliques?

Rename `max_dim` to `max_rank` in this function.

PR #337, opened 2024-02-13
Activate more ruff rules

ffl096:
This pull request activates some additional linting rules and fixes occurring errors as a result of them.

Some noticeable new rules:
- `zip()`: Use the optional `strict` parameter to error when the provided iterables are not of same length. This should catch some erroneous user inputs.
- Remove unnecessary `elif` and `else` branches if the previous branch is guaranteed to return. This reduces indentation and makes the code cleaner.
- Use list comprehensions instead of for-loops to transform lists. [This is more performant](https://docs.astral.sh/ruff/rules/manual-list-comprehension/).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,6 +90,7 @@ select = [
     "B", # bugbear rules
     "PIE", # pie rules
     "Q", # quote rules
+    "RET", # return rules
     "SIM", # code simplifications
 ]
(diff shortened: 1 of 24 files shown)


ffl096:
Codecov complaints are about changes to lines that have been and still are uncovered by tests. No new code has been added in this pull request.

USFCA-MSDS (approved):

PR #342, opened 2024-03-06
Do not document `__init__` methods

ffl096:
The `__init__` function is already documented as part of the class docstring. This is also common practice in other libraries.

Once merged, similar changes should be applied to TopoEmbedX and TopoModelX.


gbg141 (approved):
Thank you, Florian. I will apply the same change to TopoModelX.

Issue #288, opened 2023-10-18
remove_singletons in ColoredHyperGraph throws error.


gurug-dev:
What :

remove_singletons() method of CHG does not work, throws error during creating a new CHG using restrict_to_cells() method.

When: (example below)
```
        CHG = ColoredHyperGraph([[1, 2, 3], [2, 3, 4]], ranks=2)
        CHG.add_cell([9]) #add a singleton cell
        assert CHG.singletons() == [(frozenset({9}), 0)] #works
        CHG.remove_singletons() #Throws error!
```
KeyError: 'hyperedge (frozenset({1, 2, 3}), 0) is not in the complex'
The above error is thrown during runtime of restrict_to_cells.

How to reproduce:
- Create a CHG with singleton cells.
- Verify the singleton cells using CHG.singletons()
- Try to remove singletons using CHG.remove_singletons()

Places to fix:
restrict_to_cells method of ColoredHyperGraph

USFCA-MSDS:
This is being addressed with PR #348

mhajij:
Closed with PR #348

PR #348, opened 2024-04-04
small fix to chg addressing empty graph in compute laplacian

USFCA-MSDS:
(1) The code does not handle empty chg.

(2) fixing #288


ffl096 (changes requested):
Please add a test to catch this error in the future.

USFCA-MSDS:
> Please add a test to catch this error in the future.

Now the function checks if empty then it returns the zero matrix. Should we change this behavior?

ffl096:
That's not what I meant: You want to fix an error. Please write a unit test that failed previously and is fixed now with this change. This prevents regressions in the future :)

ffl096 (approved):

PR #350, opened 2024-05-03
Fixed typos and errors in all four tutorials

rodroadl:
I fixed typos and errors in all four tutorials. Errors include missing image, tex and markdown rendering.

Also, it seems like there have been update to `adjacency_matrix` and `coadjacency_matrix`, since they have entry other than 0 and 1, for example in cell In \[09\] [03_combinatorial_complexes.ipynb](https://github.com/rodroadl/TopoNetX/blob/rodroadl-tutorials/tutorials/03_combinatorial_complexes.ipynb), `CA20` has entry 2 as rank-2 cells {1,2,3,4} and {2,4,6} are connected by two rank-0 cells {2} and {4}. So I think explanations and defintion of adjacency and coadjacency should be updated accordingly as well. Let me know your thought on this.


ninamiolane (approved):
Excellent, thank you very much!

PR #361, opened 2024-05-31
New import convention for TopoNetX

### LINKED ISSUE #335: Import Convention for TopoNetX
Since #215, classes and functions are only available with explicit imports, e.g.:
```python
from toponetx.classes import SimplicialComplex
from toponetx.transform import graph_to_clique_complex

SC1 = SimplicialComplex()
SC2 = graph_to_clique_complex(nx.erdos_renyi_graph(50, 0.3))
```

Following other popular libraries, a structure supporting the following code would be easier, I think:
```python
import toponetx as tnx

SC1 = tnx.SimplicialComplex()
SC2 = tnx.graph_to_clique_complex(nx.erdos_renyi_graph(50, 0.3))
```
We should then embrace this convention in the documentation and examples as well.

In #214, it was argued that with this structure, `from toponetx import *` will clutter the global namespace. I argue that such a statement is discouraged and shouldn't be used by the user in the first place. We can further prevent this misuse by placing an appropriate `__all__` variable in the root `__init__.py` file (see e.g. NumPy).

ffl096:
Closes #335.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,64 +85,53 @@ pre-commit install
 ## Example 1: creating a simplicial complex
 
 ```python
-from toponetx.classes import SimplicialComplex
+import toponetx as tnx
 
 # Instantiate a SimplicialComplex object with a few simplices
-
-sc = SimplicialComplex([[1, 2, 3], [2, 3, 4], [0, 1]])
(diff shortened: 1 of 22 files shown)


mhajij (approved):
really important improvement

PR #394, opened 2024-10-20
updating readme file

USFCA-MSDS:
Update the readme file, new refs. Considering the new updates of the software.


ffl096:
I have reverted the changes to the examples. Following the import convention of #335, TopoNetX should always be imported and used as `tnx` without explicit imports of classes/functions.

Issue #320, opened 2023-11-15
Package is breaking because of python version 3.12.0


devendragovil:
# What 

The package file is not building with python ver `3.12.0`. Gudhi package is throwing exceptions. 

<img width="839" alt="image" src="https://github.com/pyt-team/TopoNetX/assets/39613474/3cb986e0-4444-4f76-a730-3cb0f849adb7">

This has already been raised within the Gudhi community: 

https://github.com/GUDHI/gudhi-devel/issues/998

This is an extremely important as conda by default installs the latest version in a fresh env which is 3.12.0

# Solution

We need to change pyproject.toml to alter python requirements.

devendragovil:
@mhajij @ninamiolane @ffl096 

I have tried multiple fixes, but none seem to be working at my end. Changing pyproject.toml to enforce python version <3.12 (as highlighted in #321) is not working. Beyond that, the python documentation seems to suggest not to use the python version capping as can be seen [here](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#requires-python-upper-bounds) and [here](https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special).

PR #400, opened 2024-10-28
Test against Python 3.13

### LINKED ISSUE #399: Add Python 3.13 to Test Suite

ffl096:
Run our test suite against Python 3.13.

This pull request also mocks the call to gudhi and removes the explicit dependency. Installing Gudhi on Python 3.13 fails on Github Actions. See also #320.

Closes #399.


ffl096:
This is on hold right now. Apparently `gudhi` is not available for Python 3.13, yet.

PR #403, opened 2024-11-21
Fix from_spharapy naming in SimplicialComplex

luisfpereira:
This PR fixes a typo in `SimplicialComplex.from_spharapy` (making it consistent with `SimplicialComplex.to_spharapy` and the library name).


mhajij (approved):

ffl096 (changes requested):
Please keep the old function name as deprecated function that calls the new function. No need to break user code here :)

luisfpereira:
Thanks for finishing and merging this @ffl096. (I've completely forgotten I had it open.)

